PR #17, opened 2015-07-07
Correct string #30133 to "Post-Padding (minutes)"

redion1992:
Previously duplicated string #30132, meaning both pre- and post-padding settings showed up as "Pre-Padding (minutes)".


Jalle19:
Only `en_gb` needs to be changed, the rest are handled by Transifex. Right @alanwww1?

MartijnKaijser:
Correct. Any changes done to other languages here will be overwritten

redion1992:
Fixed, only en_gb is now modified.

Jalle19:
Can you make another PR for the Isengard branch?

PR #21, opened 2015-07-24
Update string #30133 to "Post-Padding (minutes)" (Isengard branch)

redion1992:
As requested in pull request #17 - this pull request is the same as that one, only on the Isengard branch. First commit was to update my copy to match current kodi-pvr copy, second one to implement the changes.


Jalle19:
Someone will have to update the changelog and version before we bump addons for Isengard 15.1.

redion1992:
The other thing needing doing is a pass through Transifex - not sure if that happens automatically or needs to be called manually.

Jalle19:
It's usually done before the final bump.

PR #40, opened 2016-02-25
Fixed editing recordings, and recording only new episodes

sub3:
-Fixed issue where user couldn't select to only record new episodes
-Fixed issue with updating one-off recordings
-Fixed issue where pre/post padding was not being loaded with recording list

### DIFF
--- a/src/pvrclient-nextpvr.cpp
+++ b/src/pvrclient-nextpvr.cpp
@@ -1140,6 +1140,15 @@ PVR_ERROR cPVRClientNextPVR::GetTimers(ADDON_HANDLE handle)
           tag.iMaxRecordings = atoi(pRulesNode->FirstChildElement("Keep")->FirstChild()->Value());
         }
 
+        // prevent duplicates
+        if (pRulesNode->FirstChildElement("OnlyNewEpisodes") != NULL)
+        {
+          if (strcmp(pRulesNode->FirstChildElement("OnlyNewEpisodes")->FirstChild()->Value(), "true") == 0)
+          {
+            tag.iPreventDuplicateEpisodes = 1;
+          }
+        }
+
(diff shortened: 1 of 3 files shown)


ksooo:
Could you please bump the addon version and update the changelog?

src/pvrclient-nextpvr.cpp
```diff
@@ -1199,16 +1208,31 @@ PVR_ERROR cPVRClientNextPVR::GetTimers(ADDON_HANDLE handle)
         if (pRecordingNode->FirstChildElement("recurring_parent") != NULL)
         {
           tag.iParentClientIndex = atoi(pRecordingNode->FirstChildElement("recurring_parent")->FirstChild()->Value());
-          if (tag.iTimerType == TIMER_ONCE_EPG)
-          {
-            tag.iTimerType = TIMER_ONCE_EPG_CHILD;
-          }
-          else
+          if (tag.iParentClientIndex != 0)
```
ksooo: Please use PVR_TIMER_NO_PARENT instead of a 'magic number'.

ksooo:
BTW: I reverted the Jarvis commit you did by "accident".

Please open a Jarvis PR after(!) we're done with this PR. Please, in case you are uncertain how to do this right, without accidentally pushing stuff directly to the repo, feel free to ask. I suggest reading https://github.com/xbmc/xbmc/blob/master/CONTRIBUTING.md, which is pretty much identical to how it works for pvr addons.

sub3:
I've corrected the items version/changelog, and updated to use PVR_TIMER_NO_PARENT

ksooo:
Cannot say anything about the nextpvr specific code. Otherwise it is looking good. Thanks.

PR #41, opened 2016-02-26
Fixed issues editing recordings, and selecting to record only new episodes

sub3:
-Fixed issue where user couldn't select to only record new episodes
-Fixed issue with updating one-off recordings
-Fixed issue where pre/post padding was not being loaded with recording list


ksooo:
Cannot say anything about the nextpvr specific code. Otherwise it is looking good. Thanks.

PR #42, opened 2016-02-28
Sync enum representation of char_class_e in struct traits with defini…

bkuhls:
…tion

Copied from
https://github.com/kodi-pvr/pvr.mediaportal.tvserver/pull/28
https://github.com/kodi-pvr/pvr.mediaportal.tvserver/pull/35

### DIFF
--- a/src/uri.h
+++ b/src/uri.h
@@ -21,13 +21,13 @@
 namespace uri
 {
   /// Char class.
-  enum char_class_e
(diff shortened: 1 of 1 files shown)


rbalint:
The fix is applied in the latest Debian package:
https://packages.qa.debian.org/k/kodi-pvr-nextpvr.html

Maybe it could be cherry-picked to Jarvis branch, too.

PR #64, opened 2017-07-07
Added minimalistic support for PVR Addon API 5.3.0

ksooo:
Related kodi PR: https://github.com/xbmc/xbmc/pull/12303


ksooo:
@sub3 Please check whether PVR_TIMER_TYPE_SUPPORTS_ANY_CHANNEL should be added to any of the timer types you have defined and open a follow-up PR in case. Otherwise "Any channel" will no longer be available in the timer settings' dialog channel list.

PR #66, opened 2017-09-16
Update to 5.7.0 PVR addon API to work with Kodi 18 (Leia)

mlburgett:
I realize this probably isn't ready to commit, but I wanted to see if I could jump start the process, so pvr.nextpvr can work again in the Leia daily builds.  I've tested this on the RPi2 platform, and the only issue I see is that live tv starts for a few seconds, stops to buffer, and then starts back up again.  That seems wrong.

Testing with a RPi3, I don't see the live tv behavior above. It seems to start and continue streaming fine.

### DIFF
--- a/src/pvrclient-nextpvr.cpp
+++ b/src/pvrclient-nextpvr.cpp
@@ -461,7 +461,7 @@ PVR_ERROR cPVRClientNextPVR::GetEpg(ADDON_HANDLE handle, const PVR_CHANNEL &chan
 
         broadcast.iUniqueBroadcastId  = atoi(pListingNode->FirstChildElement("id")->FirstChild()->Value());
         broadcast.strTitle            = title;
-        broadcast.iChannelNumber      = channel.iChannelNumber;
+        broadcast.iUniqueChannelId    = channel.iUniqueId;
         broadcast.startTime           = atol(start);
         broadcast.endTime             = atol(end);
         broadcast.strPlotOutline      = NULL; //unused
@@ -754,7 +754,22 @@ PVR_ERROR cPVRClientNextPVR::GetChannelGroups(ADDON_HANDLE handle, bool bRadio)
   return PVR_ERROR_NO_ERROR;
 }
 
+PVR_ERROR cPVRClientNextPVR::GetChannelStreamProperties(const PVR_CHANNEL* channel, PVR_NAMED_VALUE* props, unsigned int* prop_size)
+{
+  if (OpenLiveStream(*channel) == true)
+  {
+    if (!m_PlaybackURL.empty())
+    {
+      PVR_STRCPY(props[0].strName, PVR_STREAM_PROPERTY_STREAMURL);
+      PVR_STRCPY(props[0].strValue, m_PlaybackURL.c_str());
+      *prop_size = 1;
+      return PVR_ERROR_NO_ERROR;
+    }
+    return PVR_ERROR_UNKNOWN;
+  }
 
+  return PVR_ERROR_FAILED;
+}
 
 PVR_ERROR cPVRClientNextPVR::GetChannelGroupMembers(ADDON_HANDLE handle, const PVR_CHANNEL_GROUP &group)
 {
@@ -871,10 +886,6 @@ PVR_ERROR cPVRClientNextPVR::GetRecordings(ADDON_HANDLE handle)
         snprintf(artworkPath, sizeof(artworkPath), "http://%s:%d/service?method=recording.fanart&sid=%s&recording_id=%s", g_szHostname.c_str(), g_iPort, m_sid, tag.strRecordingId);
         PVR_STRCPY(tag.strFanartPath, artworkPath);
 
-        CStdString strStream;
-        strStream.Format("http://%s:%d/live?recording=%s", g_szHostname, g_iPort, tag.strRecordingId);
-        strncpy(tag.strStreamURL, strStream.c_str(), sizeof(tag.strStreamURL)); 
-
         /* TODO: PVR API 5.0.0: Implement this */
         tag.iChannelUid = PVR_CHANNEL_INVALID_UID;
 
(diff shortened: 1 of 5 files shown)


src/pvrclient-nextpvr.cpp
```diff
@@ -461,7 +461,7 @@ PVR_ERROR cPVRClientNextPVR::GetEpg(ADDON_HANDLE handle, const PVR_CHANNEL &chan
 
         broadcast.iUniqueBroadcastId  = atoi(pListingNode->FirstChildElement("id")->FirstChild()->Value());
         broadcast.strTitle            = title;
-        broadcast.iChannelNumber      = channel.iChannelNumber;
+		broadcast.iUniqueChannelId    = channel.iUniqueId;
```
Jalle19: Whitespace

Jalle19: It looks like you've used tabs as whitespace while the rest of the code uses two spaces. Do you think you could change that?

mlburgett:
whitespace fixed.

src/pvrclient-nextpvr.cpp
```diff
@@ -1734,13 +1752,19 @@ PVR_ERROR cPVRClientNextPVR::UpdateTimer(const PVR_TIMER &timerinfo)
...
+    XBMC->Log(LOG_NOTICE, "New channel uid equal to the already streaming channel. Skip re-tune.");
+    return true;
+  }
+	
   m_PlaybackURL = "";
+  m_iLiveStreamUID = 0;
 
   XBMC->Log(LOG_DEBUG, "OpenLiveStream(%d:%s) (oid=%d)", channelinfo.iChannelNumber, channelinfo.strChannelName, channelinfo.iUniqueId);
-  if (strstr(channelinfo.strStreamURL, "live?channel") == NULL)
+//if (strstr(channelinfo.strStreamURL, "live?channel") == NULL)
```
ksooo: Remove this, if it is no longer needed.

mlburgett: sure.  fixing whitespace errors too.

src/pvrclient-nextpvr.h
```diff
@@ -175,6 +175,7 @@ class cPVRClientNextPVR : P8PLATFORM::CThread
 
   CStdString              m_PlaybackURL;
   LiveShiftSource        *m_pLiveShiftSource;
+  unsigned int			  m_iLiveStreamUID;
```
Jalle19: Minor whitespace snafu here still

src/pvrclient-nextpvr.cpp
```diff
@@ -1026,6 +1033,17 @@ PVR_ERROR cPVRClientNextPVR::GetRecordingEdl(const PVR_RECORDING& recording, PVR
   return PVR_ERROR_FAILED;
 }
 
+PVR_ERROR cPVRClientNextPVR::GetRecordingStreamProperties(const PVR_RECORDING* recording, PVR_NAMED_VALUE* props, unsigned int* prop_size)
+{
+  char line[256];
+  sprintf(line, "http://%s:%d/live?recording=%s", g_szHostname.c_str(), g_iPort, recording->strRecordingId);
```
ksooo: Please use `snprintf` to get resistent to buffer overflows.

src/pvrclient-nextpvr.cpp
```diff
@@ -1734,13 +1752,18 @@ PVR_ERROR cPVRClientNextPVR::UpdateTimer(const PVR_TIMER &timerinfo)
 
 /************************************************************/
 /** Live stream handling */
-
 bool cPVRClientNextPVR::OpenLiveStream(const PVR_CHANNEL &channelinfo)
 {
+  if (channelinfo.iUniqueId == m_iLiveStreamUID)
+  {
+    XBMC->Log(LOG_NOTICE, "New channel uid equal to the already streaming channel. Skip re-tune.");
+    return true;
```
ksooo: Why this change? Is this a theoretical case or something you experienced?

src/pvrclient-nextpvr.cpp
```diff
@@ -1734,13 +1752,18 @@ PVR_ERROR cPVRClientNextPVR::UpdateTimer(const PVR_TIMER &timerinfo)
...
+  }
+  
   m_PlaybackURL = "";
+  m_iLiveStreamUID = 0;
 
   XBMC->Log(LOG_DEBUG, "OpenLiveStream(%d:%s) (oid=%d)", channelinfo.iChannelNumber, channelinfo.strChannelName, channelinfo.iUniqueId);
-  if (strstr(channelinfo.strStreamURL, "live?channel") == NULL)
   {
```
ksooo: If you remove the condition, please also remove the corresponding block markers (the curly braces).

src/pvrclient-nextpvr.cpp
Jalle19: I was thinking the same thing but didn't bother to comment

mlburgett: It's theoretical, given the check a few lines below that there's already a stream open, it seemed prudent._

mlburgett:
squashed.

ksooo:
Thanks.

ksooo:
I just realized that `GetChannelStreamProperties` only needs to be implemented if the addon formerly was filling member `PVR_CHANNEL::strStreamURL` in `GetChannels` which is not the case for this addon. I'm not evn sure whether it actually works if an addon implements both `OpenLiveStream` and `GetChannelStreamProperties`. 

Could you maybe come up with a PR that removes the implementation of `GetChannelStreamProperties` (incl. micro addon version bump)?

mlburgett:
It does work, I've tested this one, since it's what I use.  I'll test without it today, to make sure it continues to work.

ksooo:
> It does work, I've tested this one, since it's what I use. 

 Sure, but this might only work "by luck". So, it's better to do it like intended.

> I'll test without it today, to make sure it continues to work.

 Yeah, that would be nice. I think it will work without `GetChannelStreamProperties`.

PR #67, opened 2017-09-20
Remove unneeded code introduced in last change.

mlburgett:
Thanks to @ksooo  for pointing out that GetChannelStreamProperties() only needs to be implemented if PVR_CHANNEL::strStreamURL was being populated in GetChannels()


ksooo:
Thanks again.

PR #68, opened 2017-09-24
Implement GetChannelStreamProperties to return URL

mlburgett:
And clean up now unused code. (Including recording functions rendered obsolete by implementing GetRecordingStreamProperties)


rhocheck:
I don't think this is a good idea. When going form the specialised to the generic, you'll lose the capabilities only the specialised does provide. The ordinary http protocol is not designed for live streaming and in particular not for PVR. You can't signal to player that you open a realtime stream which will result in stalling. This is only one disadvantage.

FernetMenta:
I don't think this is a good idea. When going form the specialised to the generic, you'll lose the capabilities only the specialised does provide. The ordinary http protocol is not designed for live streaming and in particular not for PVR. You can't signal to player that you open a realtime stream which will result in stalling. This is only one disadvantage.

mlburgett:
I'm going to close this, I didn't intend to open this here yet, but I wanted to ask you about your comment.

From the discussion on the Leia API changes, I thought this was the way it was going, pvr addons were just going to provide playable URLs if they can, and OpenLiveStream/OpenRecordedStream were going away?

ksooo:
> pvr addons were just going to provide playable URLs if they can, and OpenLiveStream/OpenRecordedStream were going away?

 Nope. There are and will be several "stages", from worst to best, function-wise:

1. URLs (via Get*StreamProperties)
2. Streams (via Open*Stream + Read*Stream + ...)
3. Demuxers for LiveTV (via OpenLiveStream + DemuxRead + ...)

FernetMenta:
All streaming related functions will go away from PVR API. In this regard @mlburgett was partly correct. But inputstream API will be used instead.

ksooo:
Sure, names will change (dedicated PVR stream functions today, general inputstream functions tomorrow), but ghist stays the same. I guess, even 'tomorrow' this addon will provide it's own (input)stream implementation, which basically will be the same code as the stream implementation uses today (I guess).

mlburgett:
I must need to do more reading. I thought the pvr addons would need to provide inputstream functions only if inputstream couldn't already handle the url, but that the preferred path in pvr addon would be to produce a url that inputstream could play.  From the list above, that's the least desirable path.

PR #73, opened 2018-04-30
Enable timeshift

mlburgett:
Pulled some code from pvr.vbox as a basis for these changes, which don't
use a local file for timeshifting, but rather depend on the backend being in timeshift mode to handle it.  Implement GetStreamTimes() to enable timeshifting.


sub3:
That's some significant amount of code changes. Superficially it looks ok,  but it's not clear to me exactly how it works, so I'll have to try it here to check it actually works.

sub3:
Reported offline to mlburgett, I'd had issues compiling his changes. He has since fixed these and added the commits, and I'm happy to go ahead with merging these changes.

mlburgett:
Thank you, Sub3!

PR #75, opened 2018-06-18
Add implementation for GetReadStreamChunkSize()

mlburgett:
also, fix some cases were -1 should be returned instead of 0, when a failure occurs, when counting/listing recordings, and timers.


pvr.nextpvr/changelog.txt
```diff
@@ -1,3 +1,9 @@
+v3.3.2
+- use iUniqueId instead of iChannelNumber for live streaming.
+
+v3.3.1
+- Implent GetStreamReadChunkSize()
```
Jalle19: Typo

PR #79, opened 2019-01-04
In-progress file changes, trailing space cleanup, resume in progress …

emveepee:
…files

### DIFF
--- a/src/buffers/RecordingBuffer.cpp
+++ b/src/buffers/RecordingBuffer.cpp
@@ -21,6 +21,8 @@
 
 #include "RecordingBuffer.h"
 
+#include "libXBMC_pvr.h"
+
 using namespace timeshift;
 
 PVR_ERROR RecordingBuffer::GetStreamTimes(PVR_STREAM_TIMES *stimes)
@@ -28,7 +30,7 @@ PVR_ERROR RecordingBuffer::GetStreamTimes(PVR_STREAM_TIMES *stimes)
   stimes->startTime = 0;
   stimes->ptsStart = 0;
   stimes->ptsBegin = 0;
-  stimes->ptsEnd = ((int64_t )m_Duration) * DVD_TIME_BASE;
+  stimes->ptsEnd = ((int64_t ) RecordingBuffer::Duration() ) * DVD_TIME_BASE;
   XBMC->Log(LOG_DEBUG, "RecordingBuffer::GetStreamTimes called!");
   return PVR_ERROR_NO_ERROR;
 }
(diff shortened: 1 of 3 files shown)


src/buffers/RecordingBuffer.cpp
```diff
@@ -21,14 +21,16 @@
 
 #include "RecordingBuffer.h"
 
+#include "libXBMC_pvr.h"
```
mlburgett: This shouldn't be needed.  RecordingBuffer.h includes Buffer.h, which includes client.h, which includes this file.

src/buffers/RecordingBuffer.h
```diff
@@ -76,5 +72,10 @@ namespace timeshift {
     
     void SetDuration(int duration) { m_Duration = duration; XBMC->Log(LOG_ERROR, "XXXXX Duration set to %d XXXXX", m_Duration); }
     int GetDuration(void) { return m_Duration; XBMC->Log(LOG_ERROR, "XXXXX Duration set to %d XXXXX", m_Duration); }
+
+    bool Open(const std::string inputUrl,const PVR_RECORDING &recording);
+    int Read(byte *buffer, size_t length);
```
mlburgett: Using 'override' signals the compiler that the function should overload a virtual function in the base, and will generate a warning if it doesn't.   I'd recommend removing it from prototypes unless there's a compelling reason.

src/pvrclient-nextpvr.cpp
```diff
@@ -34,10 +34,10 @@
 #include "tinyxml.h"
 
 #if defined(TARGET_WINDOWS)
-  #define atoll(S) _atoi64(S)
```
mlburgett: Not a fan of gratuitous white space changes mixed with code changes.

emveepee: I am not a fan of white space being there in the first place but I will revert these

src/buffers/RecordingBuffer.h
emveepee: Open needs to be new because of taking two variables.

Read needs an override and I changed it to match TimeShiftBuffer.h

src/buffers/RecordingBuffer.cpp
emveepee: Ok removed

emveepee:
/home/jenkins/workspace/binary-addons/kodi-android-aarch64-master/tools/depends/target/binary-addons/pvr.nextpvr/src/buffers/RecordingBuffer.h:78:26: error: extra qualification on member 'Duration'

Extra use of RecordingBuffer::   removed.

PR #80, opened 2019-01-05
In-progress recording changes (attempt #2)

emveepee:


src/buffers/RecordingBuffer.cpp
```diff
@@ -42,3 +42,58 @@ time_t RecordingBuffer::GetBufferEndTime()
...
+  }
+}
+bool RecordingBuffer::Open(const std::string inputUrl,const PVR_RECORDING &recording)
+{
+  m_Duration = recording.iDuration;
+  if (recording.iDuration + recording.recordingTime > time(nullptr))
+  {
+    m_startTime = recording.recordingTime;
+    XBMC->Log(LOG_DEBUG, "RecordingBuffer::Open In Progress %d %lld", recording.iDuration, recording.recordingTime);
+    m_isRecording = TRUE;
```
mlburgett: Checks are failing right now, because of these.  use true and false, not TRUE and FALSE in C++

emveepee:
Sorry I have to do another build this failed on TRUE and FALSE not being defined on linux.  Changing to

PR #81, opened 2019-01-05
In-progress  recording improvements (attempt #3)

emveepee:


mlburgett:
Needs version bump, and changelog.txt update.  3.3.9 is already in (at least some) nightlies.

emveepee:
> 
> 
> Needs version bump, and changelog.txt update. 3.3.9 is already in (at least some) nightlies.

Sub never tagged 3.3.9-Leia so it should not have make it to the nightlies unless they are created manually.  I can do this if you'd like.

mlburgett:
Please do.  I think every change deserves a version #.  (They're cheap)

emveepee:
Ok I need feedback for sub on the version on whether we bump to 3.4.0 or not

mlburgett:
I would expect it's 3.3.10.

emveepee:
Ok I will build that now.  I have another small PR ready and waiting that allows playback of via the filesystem instead of http so we can figure out the version jump at that time.

PR #82, opened 2019-01-06
More fixes for live tv tsb and seeks

mlburgett:
Rework time/buffer housekeeping so we know (approximately where the start/end of the tsb is.

Also, remove some outdated functions.


mlburgett:
Not sure what you're referring to in the code, but on the client side of the socket, the slip file(s) are presented as a single virtual file.  All the issues about which bits are in which slip file are hidden from the client side.  @sub3, do you see an issue with how g_timeShiftBufferSeconds is used?

sub3:
I wont pretend to have a full understanding of how the everything works in these newer versions of the NextPVR addon. I've been lucky enough to be able to spend my time on other stuff.

From a live tv client perspective though - the device slip buffer size on the server is 20 minutes. In the first 20 minutes of live tv, the size of the available slip data will be the number of seconds that live tv has been running. After 20 minutes, at least 20 minutes of data will be available to the client at all times. At times there is more data available, but this is hidden from the client. 

ie, the slip buffer is 1200 seconds (20*60). The server implements this as a rolling set of four files, each containing up to 400 seconds. 3 files of 400 seconds = 1200, and the fourth most recent files is always somewhere between 0 and 400 seconds as it grows to completion. When teh fourth files hits 400 seconds, a new file is created, and the oldest deleted.

sub3:
I guess the issue is really how to translate "seconds" into "seek offset", since there is no super easy way for the client to know precisely what time represents which portion of the virtual file, and it's probably easy for it to guess at a bitrate and could guess for too far back, into an area that the application no longer has. 

I'm not 100% what it does in this scenario - as Martin says, it might currently not seek at all, rather than just go back as far as it can.

emveepee:
If my logic is wrong and we want to be accurate right now It is possible to get the slip file for epg mode with method=channel.stream.info and the sid.  If that could be extended to regular live tv it could work.  The offset value is actually better because the addon doesn't implement the Kodi time skipping option

mlburgett:
@Sub3, that sounds reasonable.  There's always some risk of trying to jump back too far, since we don't know exactly where the oldest data available is.   I'm using a bytes per second average created over the life of the session, and assuming the back end will always have at least slip seconds worth of data available (once slip seconds has elapsed.)  The actual seeking is accomplished by calculating a 32768 byte aligned block to request that (assuming it's within bounds) contains the offset from 0 being seeked to.

mlburgett:
That sounds like it could be helpful.

emveepee:
I looked over the code and 

1. my major concern is I don't understand the need for TSBTimerProc() in it's own thread.  We already have a control function GetStreamTimes() that Kodi polls every second paused or not.  Right now some checks are in GetStartTime(); but  the "timer" proc could be called in an different way

2. The code cleanup for removing unnecessary functions is good but it would have been nice to see it in a PR of its own for easy acceptance.  

3. The cleanup PR could also include the virtual PauseStream() function with no change to TimeShift and session files except for the base class needed to support future Pausing.logic changes.  Again an easy acceptance

4 Minor point but In my testing because of time needed to setup a tuner a recording the buffer actually starts here https://github.com/kodi-pvr/pvr.nextpvr/blob/master/src/buffers/TimeshiftBuffer.cpp#L137
There can be can be 2-8 seconds lost in my setup more but IPTV users can lose more.
.

mlburgett:
In my testing GetStreamTimes() was call ~3 times a second.  Since we don't really need/want other that per second resolution, this appears to be unnecessary work, so now GetStreamTimes() can just return a values  already calculated.

.

mlburgett:
Thanks for the merge, sub3!

PR #83, opened 2019-01-09
EPG based live TV, playback from files

emveepee:
1. support for EPG Based live TV.   Seek looks to be broken maybe because NextPVR is not providing a Content-Length.  Worthwhile for longer pausing
2.  Recordings will play from file system for local files, shares, and using advancedsettings.xml subtitution.  
3. DoRequest moved to a Class so the buffer classes can use it.
4. Fixed radio playback in Timeshift mode, URL was not compatible with the new buffer
5. Debug log spam reduction in Recordings


emveepee:
Closing to rewrite for file based http for epg base

PR #86, opened 2019-02-27
Add extended timeshift and several fixes

emveepee:
Fixed radio streaming, Improved radio integration, align with v18 API, add recent fixes, add wol


emveepee:
I had some platform compile issues but I believe they were all squashed into the second last commit.  The failure look like a process failure getting origin.  Not sure where to go from here?

emveepee:
I will close and resubmit a new PR from the same state.

PR #87, opened 2019-02-28
Add extended timeshift and several fixes 

emveepee:


sub3:
As discussed offline, I've merging this now that NextPVR 4.2.4 is available.

PR #89, opened 2019-03-14
Style Cleanup

emveepee:
Remove trailing spaces and tabs.

### DIFF
--- a/pvr.nextpvr/changelog.txt
+++ b/pvr.nextpvr/changelog.txt
@@ -51,7 +51,7 @@ v3.2.7
 v3.2.6
 - Updated to PVR addon API 5.9.0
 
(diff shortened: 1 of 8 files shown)


AlwinEsch:
Always good to have the code cleaner, Thanks!

emveepee:
@mlburgett  I did not include Timeshift.h  It has a few trailing spaces that I could add as an amend if that is ok with you.

sub3:
Thanks for the cleanup!

sub3:
@ksooo hopefully that merge went the way you wanted it? Unfortunately after this merge we did realise there is a 'merge and rebase' option, which I should have used? Sorry - I know just enough to be dangerous 😀

PR #90, opened 2019-03-16
Delete recording trigger change,  cleanup initialization code

emveepee:
Removed unnecessary trigger  after deleting a recording,  prevent delete of in progress recording, clean up error message.  Added playback of plugin:/ URLs.  Cleaned up MAC address logic and skip WOL for localhost.  Fixed default for settings.xml

### DIFF
--- a/pvr.nextpvr/changelog.txt
+++ b/pvr.nextpvr/changelog.txt
@@ -1,8 +1,11 @@
 v3.3.14
-- Fixed issue with meta art not showing for deleted channels
-- New settings option to enabled Kodi style recordings - defaults to NextPVR style
+- Fixed issue with art not showing up on recordings from deleted channels
+- New settings option to enabled Kodi style recordings - defaults to NextPVR style (no title, with S/E )
 - Reduce channel tuning time in Extended Timeshift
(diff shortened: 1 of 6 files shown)


emveepee:
Note that with change if a user tries to delete a folder with an in-progress recording they will see this dialog.
error

The log will show " Add-on 'NextPVR  (xx.xx.xx.xx):connected' returned an error: recording already running"

No way with Kodi avoiding an error,  we probably need to raise a Kodi issue not to delete folders with in-progress recordings.

emveepee:
I resubmitted this to deal with an incorrect default in settings.xml.  In compiling there was a failure

D:\jenkins\workspace\binary-addons\kodi-windows-x86_64-master\cmake\addons\build\p8-platform\src\p8-platform\src\windows\dlfcn-win32.cpp(276): fatal error C1090: PDB API call failed, error code '23': (0x000006BA)

NMAKE : fatal error U1077: 'C:\PROGRA~2\MIB055~1\2017\COMMUN~1\VC\Tools\MSVC\1416~1.270\bin\Hostx64\x64\cl.exe' : return code '0x2'

Outside my control.  Not sure what the next step is. Please advise

PR #91, opened 2019-03-25
Handle empty MAC address

emveepee:
NextPVR can return setting with no ServerMAC when networking is not configured.


sub3:
Thanks for fixing this.

PR #93, opened 2019-04-13
V3.3.15

emveepee:
Initial changes for NextPVR v5 (subtitle, season, episode, disable timeshift).

### DIFF
--- a/src/pvrclient-nextpvr.cpp
+++ b/src/pvrclient-nextpvr.cpp
@@ -279,6 +279,14 @@ bool cPVRClientNextPVR::Connect()
                     XBMC->QueueNotification(QUEUE_ERROR, XBMC->GetLocalizedString(30051), NEXTPVRC_MIN_VERSION_STRING);
                     return false;
                   }
+                  else if (version >= 50000)
+                  {
+                     if ( g_livestreamingmethod != RealTime)
+                     {
+                       g_livestreamingmethod = RealTime;
+                      XBMC->QueueNotification(QUEUE_ERROR,"v5 timeshifting disabled");
+                     }
+                  }
                 }
                 TiXmlElement* liveTimeshiftNode = settingsDoc.RootElement()->FirstChildElement("LiveTimeshift");
                 if (liveTimeshiftNode != NULL && g_livestreamingmethod != RealTime)
@@ -348,6 +356,7 @@ bool cPVRClientNextPVR::Connect()
             }
 
             m_bConnected = true;
+            LoadLiveStreams();
             XBMC->Log(LOG_DEBUG, "session.login successful");
             return true;
           }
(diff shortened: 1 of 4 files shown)


sub3:
Thanks Martin!

src/pvrclient-nextpvr.cpp
```diff
@@ -279,6 +279,14 @@ bool cPVRClientNextPVR::Connect()
                     XBMC->QueueNotification(QUEUE_ERROR, XBMC->GetLocalizedString(30051), NEXTPVRC_MIN_VERSION_STRING);
                     return false;
                   }
+                  else if (version >= 50000)
+                  {
+                     if ( g_livestreamingmethod != RealTime)
+                     {
+                       g_livestreamingmethod = RealTime;
+                      XBMC->QueueNotification(QUEUE_ERROR,"v5 timeshifting disabled");
```
ksooo, line 287: Wrong indentation.

src/pvrclient-nextpvr.cpp
```diff
@@ -279,6 +279,14 @@ bool cPVRClientNextPVR::Connect()
                     XBMC->QueueNotification(QUEUE_ERROR, XBMC->GetLocalizedString(30051), NEXTPVRC_MIN_VERSION_STRING);
                     return false;
                   }
+                  else if (version >= 50000)
+                  {
+                     if ( g_livestreamingmethod != RealTime)
```
ksooo, line 284: Blank between ( and  g.

src/pvrclient-nextpvr.cpp
```diff
@@ -563,17 +572,44 @@ PVR_ERROR cPVRClientNextPVR::GetEpg(ADDON_HANDLE handle, const PVR_CHANNEL &chan
...
         memset(&broadcast, 0, sizeof(EPG_TAG));
 
-        char title[128];
-        char description[1024];
-
-        strncpy(title, pListingNode->FirstChildElement("name")->FirstChild()->Value(), sizeof title);
+        string title;
+        string description;
+        string subtitle;
+        title = pListingNode->FirstChildElement("name")->FirstChild()->Value();
```
ksooo, line 578: This is not  C. Why don't you declare and init title in one line?

src/pvrclient-nextpvr.cpp
```diff
@@ -563,17 +572,44 @@ PVR_ERROR cPVRClientNextPVR::GetEpg(ADDON_HANDLE handle, const PVR_CHANNEL &chan
...
         if (pListingNode->FirstChildElement("description") != NULL && pListingNode->FirstChildElement("description")->FirstChild() != NULL)
         {
-          PVR_STRCPY(description, pListingNode->FirstChildElement("description")->FirstChild()->Value());
+          description = pListingNode->FirstChildElement("description")->FirstChild()->Value();
+        }
+        else
+        {
+          description = "";
```
ksooo, line 585: Isn't description always empty here?

src/pvrclient-nextpvr.cpp
```diff
@@ -563,17 +572,44 @@ PVR_ERROR cPVRClientNextPVR::GetEpg(ADDON_HANDLE handle, const PVR_CHANNEL &chan
...
+        }
+
+        if (pListingNode->FirstChildElement("subtitle") != NULL && pListingNode->FirstChildElement("subtitle")->FirstChild() != NULL)
+        {
+          subtitle = pListingNode->FirstChildElement("subtitle")->FirstChild()->Value();
+          if (description == subtitle + ":")
+          {
+            description = "";
```
ksooo, line 593: description.clear() is more efficient than assigning empty string.

src/pvrclient-nextpvr.cpp
```diff
@@ -563,17 +572,44 @@ PVR_ERROR cPVRClientNextPVR::GetEpg(ADDON_HANDLE handle, const PVR_CHANNEL &chan
...
+          }
+          else if (StringUtils::StartsWith(description,subtitle + ": "))
+          {
+            description = description.substr(subtitle.length()+2);
+          }
         }
         else
         {
-          description[0] = '\0';
+          subtitle = "";
```
ksooo, line 602: subtitle.clear()

src/pvrclient-nextpvr.cpp
```diff
@@ -563,17 +572,44 @@ PVR_ERROR cPVRClientNextPVR::GetEpg(ADDON_HANDLE handle, const PVR_CHANNEL &chan
...
+        }
+
+        if (pListingNode->FirstChildElement("year") != NULL && pListingNode->FirstChildElement("year")->FirstChild() != NULL)
```
ksooo, line 605: in c++, nullptr should be used instead of NULL

ksooo:
Let me know if you guys don't care about "stylistic" code enhancements, then I will  no longer look over your PRs. Personally, I think the last PRs could have gotten some more love before merge.

sub3:
Thanks ksooo - even this stylistic feedback is appreciated.  Sorry it wasn't a cleaner PR from us. When we're focusing on application function, some times we forget this type of stuff. We will work through your feedback items, and address with the next PR. Cheers.

emveepee:
Yes  I was looking for feedback, the PR got tagged a bit earlier than I expected.   I do want to move more of the old C code to C++ as I go so don't know all the conventions.   Things like NULL to nullptr will be more for a style cleanup PR.

The reason for some of the odd looking code is the API from NextPVR supports multiple clients beyond Kodi and sub didn't want to bump the server API version.  Currently the description is a combination of  "episode name: description."  Now that the episode name is coming in a separate field in v5, I have to parse it with the colon from the description.  This means at times there will be no description.

ksooo:
>  I do want to move more of the old C code to C++ as I go so don't know all the conventions. Things like NULL to nullptr will be more for a style cleanup PR.

 The second sentence contradicts the first. Either you want to make the changes as you go or with dedicated cleanup commits. ;-)

emveepee:
All the the other XML parsing code uses NULL, I thought it would be odd changing them one at a time but if that is better style I will do that in the future.

ksooo:
> All the the other XML parsing code uses NULL, I thought it would be odd changing them one at a time 

 There is no right or wrong here. For Kodi, we fix coding guideline violations whenever we touch a code line (this is how I take "as we go"). Addons can handle this differently.

 I just commented where I thought the code could be improved. It's completely up to you what you make out of this.

emveepee:
Excellent comment on fixing the code as we touch the line so I will do that going forward.  Much appreciated once again.

PR #95, opened 2019-04-25
[cmake] remove usage of kodi-platform

olafhering:
The pvr.nextpvr binary addon does not use kodi-platform.
Remove existing references from CMakeLists.txt and debian/control.


emveepee:
Is there a specific reason in Kodi that this is being requested?  Although I was not aware of this historical reference, after a bit of research I would like to make use of the features in  "util/XMLUtils.h" to clean up the tinyxml calls as I touch the code.  There may be other useful features that we are missing.

Rechi:
There is no reason to require a library which isn't used. If you want to use it eventually, you can readd it as dependency.

PR #101, opened 2019-12-15
V4.30 V5 Timeshifting

emveepee:
Implement v5 timeshifting and several improvements

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,10 +4,12 @@ project(pvr.nextpvr)
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${PROJECT_SOURCE_DIR})
 
 find_package(Kodi REQUIRED)
+find_package(kodiplatform REQUIRED)
 find_package(p8-platform REQUIRED)
 find_package(TinyXML REQUIRED)
 
 include_directories(${p8-platform_INCLUDE_DIRS}
+                    ${kodiplatform_INCLUDE_DIRS}
                     ${TINYXML_INCLUDE_DIR}
                     ${KODI_INCLUDE_DIR}/..) # Hack way with "/..", need bigger Kodi cmake rework to match right include ways
 
@@ -19,6 +21,7 @@ set(NEXTPVR_SOURCES src/client.cpp
                     src/BackendRequest.cpp
                     src/buffers/Buffer.cpp
                     src/buffers/DummyBuffer.cpp
+                    src/buffers/ClientTimeshift.cpp
                     src/buffers/TimeshiftBuffer.cpp
                     src/buffers/RecordingBuffer.cpp
                     src/buffers/CircularBuffer.cpp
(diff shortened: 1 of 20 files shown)


emveepee:
Any thoughts on this commit, I would like to backport some features to Leia to support the new version of NextPVR

sub3:
I don't have much to say on the specific code changes - I've had a quick review through it though, and it looks ok. The kodi guys may have more feedback on stylistic stuff etc. 

The testing I've done with this build appears to be working well. The timeshift skipping behaviour, particularly with the current NextPVR v5, works much better than previous versions of the addon (with previous versions of NextPVR server). Good job. 

If we don't receive any feedback from other developers in the next day, I'll merge the pull request.

phunkyfish:
Can you explain the concept of a roll for my own curiousity?

Does this timeshift implementation use some kind of ring buffer or does it grow indefinitely?

emveepee:
> 
> 
> Can you explain the concept of a roll for my own curiousity?
> 
> Does this timeshift implementation use some kind of ring buffer or does it grow indefinitely?

Sure.  I suggest that you think of it like a segmented m3u8 file where the ts segments "role" after a certain period of time.   The server guarantees that there are enough segments to buffer a user specified amount of time (default 20 minutes) even though there could be more because of the way each segment is written.

In this implementation there is no buffering in the client other than basic Kodi buffering.  GetStreamTimes() limit the start time of the buffer which is very successful.  The seek offset in Kodi is zero-based, even if segments drop off on the server but there are checks in case Kodi tries to seek too early or when the user is paused.  There is possibility a user will be paused longer then the buffer time so the internal Kodi plays briefly until fresh data is read.

In the current client version there was an attempt to move this into a complicated circular buffer in Kodi, but it wasn't 100% successful resulting in many complaints.   The logic in the new release is significantly less complicated,

phunkyfish:
Thanks for explaining, sounds cool.

I always wanted to write a local circular buffer that was useable across PVR addons. It’s a complex problem but would avoid the current disk space limitation that a lot of addons have.

emveepee:
> I always wanted to write a local circular buffer that was useable across PVR addons. It’s a complex problem but would avoid the current disk space limitation that a lot of addons have.

With our 18Mbs 1080i streams in North America any internal buffer size is always going to be an issue as.  I think live tv for most user is more chunked streaming with occasional pauses and seeks so the complexities of a circular buffer may not provide significant advantage but I will certainly try and implement an feature you provide.

sub3:
Our scheme in NextPVR works pretty well. In the early days I had originally considered implementing the timeshift buffer as a circular file, but ended up going with this rolling file scheme instead. In some ways it's a lot less complicated, and has the advantage of just making regular .ts files, that you can just take a copy of an play in any player.  

The user specifies the number of minutes they want to use for the time shift buffer, and backend starts recording, periodically switching to a new file, keeping the few most current files, and deleting the oldest file. The http streaming interface just represents this as a single contiguous stream to the client, so they don't have to deal with the complexities of it being multiple files.

phunkyfish:
I like the rolling file idea more than the circular now that you’ve explained it.

Would it be difficult to make this generic so it could be used in any addon?

Even just using a number of source files and wiring it up to the inputstream functions?

sub3:
To be honest that is probably not something we'd tackle. This is all stuff that happens in our backend, transparent to the client. We do have some historic stuff in the addon which knows about the separate files, still there to support users with the older backends, but we're moving away from that, so Kodi doesn't need to be aware of the fact there is multiple files.

phunkyfish:
Ah ok, I was thinking of something purely local on kodi. No worries.

PR #105, opened 2019-12-31
Variable initialization error

emveepee:
Variable m_complete not initialized to false, extended timeshift may not issue leases


emveepee:
I tested this fix on aarch64 and live tv in extended mode worked,  where an earlier test in 3.3.17 failed.  With initialization in other test platforms I could not duplicate the problem but m_complete not defaulting to false is a logical explanation of why leases were not being sent for this condition

https://github.com/kodi-pvr/pvr.nextpvr/blob/Matrix/src/buffers/Buffer.cpp#L87

I committed this for Leia production users first and will update Matrix after this is accepted

PR #107, opened 2020-01-02
Variable initialization error

emveepee:
Variable m_complete not initialized to false, extended timeshift may not issue leases


sub3:
Thanks

PR #108, opened 2020-01-17
Transcoded Live TV support

emveepee:
Play backend transcoded m3u8 files using the GetChannelStreamProperties()  interface

### DIFF
--- a/src/buffers/TranscodedBuffer.cpp
+++ b/src/buffers/TranscodedBuffer.cpp
@@ -1,0 +1,127 @@
+/*
+*      Copyright (C) 2020
+*
+*  This Program is free software; you can redistribute it and/or modify
+*  it under the terms of the GNU General Public License as published by
+*  the Free Software Foundation; either version 2, or (at your option)
+*  any later version.
+*
+*  This Program is distributed in the hope that it will be useful,
+*  but WITHOUT ANY WARRANTY; without even the implied warranty of
+*  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+*  GNU General Public License for more details.
+*
+*  You should have received a copy of the GNU General Public License
+*  along with XBMC; see the file COPYING.  If not, write to
+*  the Free Software Foundation, 51 Franklin Street, Fifth Floor, Boston,
+*  MA 02110-1301  USA
+*  http://www.gnu.org/copyleft/gpl.html
+*
+*/
+
+#include "TranscodedBuffer.h"
+#include "tinyxml.h"
+#include "kodi/util/XMLUtils.h"
+
+using namespace timeshift;
+using namespace ADDON;
+
+bool TranscodedBuffer::Open(const std::string inputUrl)
+{
+  if (m_channel_id != 0)
+  {
+    if (m_active)
+    {
+      SLEEP(1000);
+      Close();
+    }
+    XBMC->Log(LOG_DEBUG, "%s:%d:", __FUNCTION__, __LINE__);
+    std::string response;
+    std::string formattedRequest = "/services/service?method=channel.transcode.initiate&force=true&channel_id=" + std::to_string(m_channel_id) + m_profile.str();
+    if (NextPVR::m_backEnd->DoRequest( formattedRequest.c_str(), response) != HTTP_OK)
+    {
+      return false;
+    }
+    int status;
+    do
+    {
+      status = TranscodeStatus();
+      SLEEP(1000);
+    } while (status >= 0 && status < 100);
+
+    if (status == 100)
+    {
+      m_active = true;
+      m_nextLease = 0;
+      m_nextStreamInfo = INT64_MAX;
+      m_nextRoll = INT64_MAX;
+      m_isLeaseRunning = true;
+      m_complete = false;
+      m_leaseThread = std::thread([this]()
+      {
+        LeaseWorker();
+      });
+      return true;
+    }
+  }
+  return false;
+}
+
+void TranscodedBuffer::Close()
+{
+  if (m_active)
+  {
+    m_active = false;
+    m_complete = true;
+    m_isLeaseRunning = false;
+    if (m_leaseThread.joinable()){
+      m_leaseThread.detach();
+      XBMC->Log(LOG_DEBUG, "%s:%d: %d", __FUNCTION__, __LINE__,m_leaseThread.joinable());
+    }
+    std::string response;
+    NextPVR::m_backEnd->DoRequest("/services/service?method=channel.transcode.stop", response);
+  }
+}
+
+int TranscodedBuffer::TranscodeStatus()
+{
+  int percentage = -1;
+  std::string response;
+  if (NextPVR::m_backEnd->DoRequest("/services/service?method=channel.transcode.status", response) == HTTP_OK)
+  {
+    TiXmlDocument doc;
+    if (doc.Parse(response.c_str()) != NULL)
+    {
+      TiXmlElement* rspNode = doc.FirstChildElement("rsp");
+      if (rspNode != NULL)
+      {
+        bool final;
+        XMLUtils::GetInt(rspNode,"percentage",percentage);
+        XMLUtils::GetBoolean(rspNode,"final",final);
+        if (final)
+        {
+          XBMC->Log(LOG_DEBUG, "%s:%d: %s", __FUNCTION__, __LINE__,response.c_str());
+          if (percentage != 100)
+            percentage = -1;
+        }
+      }
+    }
+  }
+
+  return percentage;
+}
+
+int TranscodedBuffer::Lease()
+{
+  XBMC->Log(LOG_DEBUG, "%s:%d:", __FUNCTION__, __LINE__);
+  m_nextStreamInfo = time(nullptr) + 5;
+  return true;
+}
+
+bool TranscodedBuffer::GetStreamInfo()
+{
+  /* should only be called at exit*/
+  XBMC->Log(LOG_DEBUG, "%s:%d: %d", __FUNCTION__, __LINE__,m_nextStreamInfo);
+  Close();
+  return true;
+}
--- a/src/buffers/TranscodedBuffer.h
+++ b/src/buffers/TranscodedBuffer.h
@@ -1,0 +1,57 @@
+#pragma once
+/*
+*      Copyright (C) 2020
+*
+*  This Program is free software; you can redistribute it and/or modify
+*  it under the terms of the GNU General Public License as published by
+*  the Free Software Foundation; either version 2, or (at your option)
+*  any later version.
+*
+*  This Program is distributed in the hope that it will be useful,
+*  but WITHOUT ANY WARRANTY; without even the implied warranty of
+*  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+*  GNU General Public License for more details.
+*
+*  You should have received a copy of the GNU General Public License
+*  along with XBMC; see the file COPYING.  If not, write to
+*  the Free Software Foundation, 51 Franklin Street, Fifth Floor, Boston,
+*  MA 02110-1301  USA
+*  http://www.gnu.org/copyleft/gpl.html
+*
+*/
+
+#include  "../BackendRequest.h"
+#include "DummyBuffer.h"
+#include <sstream>
+
+namespace timeshift {
+
+  class TranscodedBuffer : public DummyBuffer
+  {
+  public:
+    TranscodedBuffer() : DummyBuffer()
+    {
+      m_profile << "&profile=" << g_iResolution << "p+-+" << g_iBitrate << "Kbps";
+      XBMC->Log(LOG_NOTICE, "TranscodedBuffer created! %s", m_profile.str().c_str());
+    }
+
+    ~TranscodedBuffer() {}
+
+    bool Open(const std::string inputUrl);
+
+    void Close();
+
+    int TranscodeStatus();
+
+    int Lease();
+
+    bool CheckStatus();
+
+    bool GetStreamInfo();
+
+  private:
+    std::ostringstream m_profile;
+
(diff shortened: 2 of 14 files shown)


src/pvrclient-nextpvr.cpp
```diff
@@ -285,17 +285,31 @@ bool cPVRClientNextPVR::Connect()
...
+                        XBMC->QueueNotification(QUEUE_ERROR,"NextPVR v5.0.1+ required for timeshift");
                       }
                     }
                   }
+                  else
+                  {
+                    if (g_livestreamingmethod == Transcoded)
+                    {
+                        g_livestreamingmethod = RealTime;
+                        XBMC->QueueNotification(QUEUE_ERROR,"NextPVR v5 required for transcoding");
```
ksooo: localization?

src/pvrclient-nextpvr.cpp
```diff
@@ -285,17 +285,31 @@ bool cPVRClientNextPVR::Connect()
...
+                      }
+                      else if (version >= 50001)
                       {
                         g_livestreamingmethod = ClientTimeshift;
                       }
                       else
                       {
                         g_livestreamingmethod = RealTime;
-                        XBMC->QueueNotification(QUEUE_ERROR,"Upgrade to NextPVR version 5.0.1 or higher");
+                        XBMC->QueueNotification(QUEUE_ERROR,"NextPVR v5.0.1+ required for timeshift");
```
ksooo: localization?

src/buffers/TranscodedBuffer.h
```diff
@@ -0,0 +1,58 @@
...
+*  the Free Software Foundation, 51 Franklin Street, Fifth Floor, Boston,
+*  MA 02110-1301  USA
+*  http://www.gnu.org/copyleft/gpl.html
+*
+*/
+
+#include  "../BackendRequest.h"
+#include "DummyBuffer.h"
+#include <sstream>
+using namespace ADDON;
```
ksooo: `using namespace` in headers is really very bad coding style

src/pvrclient-nextpvr.cpp
```diff
@@ -285,17 +285,31 @@ bool cPVRClientNextPVR::Connect()
                   {
                     if ( g_livestreamingmethod != RealTime)
                     {
-                      if (version >= 50001)
+                      if (g_livestreamingmethod == Transcoded)
+                      {
+                        delete m_timeshiftBuffer;
+                        m_timeshiftBuffer = new timeshift::TranscodedBuffer();
```
ksooo, line 294: is it guaranteed that m_timeshiftBuffer will not leak?

src/buffers/TranscodedBuffer.cpp
```diff
@@ -0,0 +1,129 @@
...
+    m_complete = true;
+    m_isLeaseRunning = false;
+    if (m_leaseThread.joinable()){
+      m_leaseThread.detach();
+      XBMC->Log(LOG_DEBUG, "%s:%d: %d", __FUNCTION__, __LINE__,m_leaseThread.joinable());
+    }
+    std::string response;
+    if (NextPVR::m_backEnd->DoRequest("/services/service?method=channel.transcode.stop", response) != HTTP_OK)
+    {
+    }
```
ksooo: empty `if`body makes no sense. missing error handling?

src/pvrclient-nextpvr.cpp
emveepee: The NextPVR backend setup is currently English only and only English forum support is provided.  I will create an issue to localize these  QueueNotifications along with the existing ones.  We might need to add extra logging to help identify these errors if a user comes to the forum asking what the message means

src/buffers/TranscodedBuffer.h
emveepee: The buffer class design originated with another user in 2018 https://github.com/kodi-pvr/pvr.nextpvr/commit/3b5b0a89e6276b67e2eca2837e910b8daec80611#diff-fcaaa45c893f266e0a2dac97e871ae8d but I can remove it from this class and modify other classes as I touch them.

src/pvrclient-nextpvr.cpp
emveepee: Again this goes back to the 2018 class design, with the application never anticipating a null return from any buffer constructor.  So far this hasn't been a reported issue,  but I will test this when these buffers are first called in OpenLiveStream() for existing code and GetStreamProperties() for this PR

src/buffers/TranscodedBuffer.cpp
emveepee: The error handling is in DoRequest().  The backend is closed source and the API is not documented so it is difficult to know how to handle any error.  Since the backend will automatically stop transcoding once the heartbeat lease stops being sent I can remove this.

emveepee:
@phunkyfish  could you please review the GetChannelStreamProperties() code https://github.com/emveepee/pvr.nextpvr/blob/live-transcoding/src/pvrclient-nextpvr.cpp#L997  Is this is the proper way to open an m3u8 URL with your recent changes?  Also do you know if is there a way to have a seekable m3u8 file for LiveTV?  I mark it real time for know but perhaps this can be improved.  

Thanks

src/buffers/TranscodedBuffer.cpp
ksooo: Empty body makes no sense.

src/pvrclient-nextpvr.cpp
ksooo: Don't get how localization of the add-on  should be related to the backend as the add-on creatEs the UI string here. You do not pass something you get from the backend to Kodi. So, add a new string to en_gb strings.po and that's it.

src/buffers/TranscodedBuffer.h
ksooo: Please do. We should not replicate old bugs when we copy old code.

src/pvrclient-nextpvr.cpp
ksooo: For me this looks like a memory leak. Allocation on connect. Nothing on disconnect.

emveepee: The default base classes are instantiated in the constructor and what your are seeing is a class morphing into a new type from the default.  There is a leak in that they aren't deleted in the destructor but the addon code tells the user to restart anyway.

src/pvrclient-nextpvr.cpp
emveepee: In the amended update I was able to map all notifications to existing localization

src/pvrclient-nextpvr.cpp
emveepee: See above

src/buffers/TranscodedBuffer.h
emveepee: I removed this from the classes I touched in the PR

src/pvrclient-nextpvr.cpp
emveepee: Classes are now deleted in the Destructor()

src/buffers/TranscodedBuffer.cpp
emveepee: Removed

phunkyfish:
You'll need to more specific on recent changes? Sorry a little jet lagged ;)

Do you mean:

1. Passing an m3u8 to kodi's internal libavformat?
2. Playing LiveTv from an EPG tag? 
3. Or the (very) new inputstream? 

Maybe just describe what it is you want to do.

For the new inputstream the hope is to extend it down the line to support timeshift. Right now it can only support pseudo-timeshift on streams that support catchup.

emveepee:
> 
> 
> You'll need to more specific on recent changes? Sorry a little jet lagged ;)
> 

The NextPVR backend uses HLS m3u8 as the format for server-side transcoded streaming.  Direct play is working fine with mpegts format but I wanted the addon to support both methods and  I don't want to incorporate a full m3u8 reader.

I original hoped to open an inputstream to the m3u8 file in OpenLiveStream()  in real time play but I could not figure out how I could tell Kodi to reload the m3u8 file as it grew and read the new ts streams

What I have done and is working,  with limitations, is switching modes to use GetChannelStreamProperties() to stream m3u8 files.   The link I pointed to is the link I provided is the code for creating the real time link.   My question was whether or not it would be possible to tell the video player through GetChannelStreamProperties() that the m3u8 file was seekable?

Regarding limitations, it looks like GetChannelStreamProperties() blocks the UI thread so I can't display any messages to the user during the preparation time that is required for the m3u8 to be ready.  Not only that the mouse and key board are unresponsive.

The other limitation is on a channel switch without stop Kodi continues to request the original m3u8 file after the switch is requested and since the addon needs to prepare the same URL for the new channel, the backend gets confused on the request.

phunkyfish:
Got you. Then what you have now is the best that can be done currently.

Once I get around to extending the new inputstream addon to support timeshift (if it’s possible) you could potentially use that down the line.

emveepee:
> 

> Once I get around to extending the new inputstream addon to support timeshift (if it’s possible) you could potentially use that down the line.

OK great I will be happy to help you test that, this PR code is really best using OpenLiveStream() and it would probably fix both issues too.

emveepee:
@phunkyfish if we move to inputstream.ffmpegdirect can an m3u8 file be called from OpenLiveStream() or still from GetChannelStreamProperties?  Also NextPVR m3u8 files are seekable but the time base is the beginning of the m3u8 file which is not the same as any start time.  How do we mark these URL's seekable?  Thanks for the work on this.

phunkyfish:
Everything with an inputstream addon is via `GetChannelStreamProperties` and `GetEPGTagStreamProperties`. There is no notion of OpenLiveStream but I would like to add callbacks so the PVR addons can be notified if a stream is opened successfully and when it's closed. Moving away from `OpenLiveStream()` and PVR addons internal demuxers is important as that limits PVR to one stream at a time meaning concepts like Picture in Picture etc. are not a possibility.

Right now if you just hook up the new addon as is and just pass a stream URL you should get the same as what kodi gives you, playback with no ability to seek. If you use catchup mode then you can tell the addon about the beginning of the catchup window, end etc. and then the URL is modified to support seeking (essentially a SEEK_SET from video player). The format of the catchup URL is entirely up to you. 

For example: `http://mysite.com/streamX?cutv={Y}-{m}-{d}T{H}:{M}:{S}`
Just lets you set the Year, Month, Day, Hour, Minute and seconds where to play from. 

Similarly: `http://mysite.com/streamX?cutv={utv}` Just lets you set the utc timestamp where to start playback from. 

Similarly: `http://mysite.com/streamX?cutv={S}` Just lets you set the seconds where to start playback from. 

It's pretty flexible and should work for almost any backend I think. Does this fit you use case?

BTW, there is a good chance you will find some issues with the new addon, it could use some field testing ;)

emveepee:
Our current use case is seekable m3u8 files.   Live TV needs to handle long playback time, perhaps all day, with transitions from show to show, I am not sure if that is in the catchup use case.  As a result,
after a client determined buffer time when new ts files are queued to the playlist the earlier ts files will be dequeued so our start "time" is not fixed.    Hopefully that is not an issue.

I do like the idea of Open and Close callback.  I see Open more than a simple notification,  GetChannelStreamProperties() is blocking and stalls Kodi quite hard so I would like to see the Open call fix that Open could signal back to you that the backend is ready for streaming.  I takes time for the backend to generate that first ts file in the playlist, which is different then the IPTV model.

Also currently when a user requests a channel change, GetChannelStreamProperties  is called but  at the same time Kodi continues to try and stream the "dead" m3u8 file.  Hopefully Close and Open can fix that.

What would be nice is a callback telling us that the inputstream is streaming which we need send for keepalives.  I use SignalStatus for that but it is not PIP aware either and it would not be my first choice.   Something like IsRealtime and the (removed) IsTimeshifting callback would make sense.

I will make the changes to use your new method and certainly the  small number of users using the new feature can help test this for both of us.

phunkyfish:
Cool, so currently catch up uses a moving window for Live TV. I.e it's 3 days and moves as time progresses. What you are saying is that you start and then the end time increases but start time can stay the same? If so this is a reasonable extension of the addon that makes sense.

We also plan to extend the the inputstream so you can send updated stream properties as well as the callbacks. That should solve for the dead M3U8 file problem I think.

Really where we want to get to is that any logic related to seeking the stream is generic and in the inputstream addon and then occasionally we send the inputstream addon some extra information for any actions it needs to take (such as keep alives or new programme information, durations etc). But really the end goal is to be able to move any complex inputstream mgmt to the new addon and make the PVR addons simpler (hopefully!).

emveepee:
For start time it depends on which time you are meaning.  In the PVR world we have ptsStart and also ptsBegin and ptsBegin is the important one because it limits the scroll back and they can diverge.

In an m3u8 using EXTINF,1 as an example  the m3u8 file might start off 1.ts, 2.ts etc but eventually the first file of m3u8 will be 2.ts then 3.ts etc so the earliest point a user could seek does change.

I have reservations of channel changes being property changes but I will wait and see how that works.

phunkyfish:
So all the ptsStart, ptsEnd etc move to the inputstream addon. You simply construct a start URL and a URL with placeholders (used for seeking) to the addon. These time based placeholders get updated on each seek. The logic we would be missing would be on how the seek window logic gets updated, I.e. the equivalent of `GetStreamTimes()`.

Channel changes on property updates would be optional. You can leverage them or not depending on your use case.

emveepee:
> 
> It's a fairly minor thing, but I do wonder if it should have used "profile" instead of "resolution" for the wording, since it'd more closely match the meaning and naming used in the NextPVR backend? If it says 'resolution' some users will probably wonder 'where do I control the bitrate?'.

Makes sense I will change it to match the backend and I maybe it can be expanded upon if we move to v19 style xml setting.  The mix and match of v4 and v5 settings is already a bit confusing and I'd like to rework it.

PR #111, opened 2020-02-29
Enable seeking while paused

emveepee:
Kodi reads live TV stream while paused, need to keep stream open to allow seeking.


emveepee:
Since there were no other comments  I have added const and amended the PR.  If there are no other comments I  would like this merged and tagged and hopefully this can make it the full distro builds along with 18.6.

phunkyfish:
Would you consider this a fix or a feature?

If it’s a feature then increment the minor version instead. 

It sounds like a feature to me.

emveepee:
> 
> 
> Would you consider this a fix or a feature?
> 
It's a fix in that Kodi is crashing because the backend doesn't like not being able to read while paused and attempts a skip.  For all users there are 100's of thousands of reads going on while paused even when not skipping, I assume it is trying to buffer.

> If it’s a feature then increment the minor version instead.
> 
> It sounds like a feature to me.

I don't know is there a guide on on when to do this?

phunkyfish:
If it’s a fix then changing the micro number is correct so all correct.

phunkyfish (approved):
LGTM

PR #112, opened 2020-03-01
Enable seeking while paused 

emveepee:
Kodi reads live TV stream while paused, need to keep stream open to allow seeking and avoid crash


phunkyfish:
This one is just missing the addon xml and changelog

emveepee:
> 
> 
> This one is just missing the addon xml and changelog

Are you suggesting that we need a changelog and version for every PR? @AlwinEsch created 4.3.1 without a changelog and I was piggy backing off that "version" for when we tag   I have another change planned I was going to update the changelog at that time.

emveepee:
Oops clicked close and comment in error.

phunkyfish:
I guess once the changelog is updated per release that’s fine.

@AlwinEsch maintains all binary addons so generally we regard him as having superpowers 😃

phunkyfish (approved):
LGTM

PR #115, opened 2020-03-19
Changes for Matrix API changes

emveepee:
Cache EPG event ID to make system usable after daily update.  Lose link to recordings.

### DIFF
--- a/src/pvrclient-nextpvr.cpp
+++ b/src/pvrclient-nextpvr.cpp
@@ -613,11 +613,8 @@ PVR_ERROR cPVRClientNextPVR::GetEpg(ADDON_HANDLE handle, int iChannelUid, time_t
         broadcast = {};
 
         string title;
-        title.clear();
         string description;
-        description.clear();
         string subtitle;
-        subtitle.clear();
         XMLUtils::GetString(pListingNode,"name",title);
         XMLUtils::GetString(pListingNode,"description",description);
 
@@ -694,25 +691,32 @@ PVR_ERROR cPVRClientNextPVR::GetEpg(ADDON_HANDLE handle, int iChannelUid, time_t
         broadcast.iEpisodePartNumber = EPG_TAG_INVALID_SERIES_EPISODE;
 
         std::string original;
-        original.clear();
         XMLUtils::GetString(pListingNode,"original",original);
         broadcast.strFirstAired = original.c_str();
 
-        std::string significance;
-        significance.clear();
-        XMLUtils::GetString(pListingNode,"significance",significance);
-        if (significance == "Live")
+        bool firstrun;
+        if (XMLUtils::GetBoolean(pListingNode,"firstrun",firstrun))
         {
-          broadcast.iFlags = EPG_TAG_FLAG_IS_NEW;
-        }
-
-        if (m_showNew)
-        {
-          bool firstrun;
-          if (XMLUtils::GetBoolean(pListingNode,"firstrun",firstrun))
+          if (firstrun)
           {
-            if (firstrun)
+            std::string significance;
+            XMLUtils::GetString(pListingNode,"significance",significance);
+            if (significance == "Live")
+            {
+              broadcast.iFlags = EPG_TAG_FLAG_IS_LIVE;
+            }
+            else if (significance.find("Premiere") != string::npos)
+            {
+              broadcast.iFlags = EPG_TAG_FLAG_IS_PREMIERE;
+            }
+            else if (significance.find("Finale") != string::npos)
+            {
+              broadcast.iFlags = EPG_TAG_FLAG_IS_FINALE;
+            }
+            else if (m_showNew)
+            {
               broadcast.iFlags = EPG_TAG_FLAG_IS_NEW;
+            }
           }
         }
         PVR->TransferEpgEntry(handle, &broadcast);
@@ -1293,6 +1297,17 @@ bool cPVRClientNextPVR::UpdatePvrRecording(TiXmlElement* pRecordingNode, PVR_REC
     tag->iGenreType = EPG_GENRE_USE_STRING;
     tag->iGenreSubType = 0;
     PVR_STRCPY(tag->strGenreDescription,genres.c_str());
+  }
+
+  std::string significance;
+  XMLUtils::GetString(pRecordingNode,"significance",significance);
+  if (significance.find("Premiere") != string::npos)
+  {
+    tag->iFlags = PVR_RECORDING_FLAG_IS_PREMIERE;
+  }
+  else if (significance.find("Finale") != string::npos)
+  {
+    tag->iFlags = PVR_RECORDING_FLAG_IS_FINALE;
   }
 
   return true;
--- a/src/pvrclient-nextpvr.cpp
+++ b/src/pvrclient-nextpvr.cpp
@@ -159,6 +159,10 @@ cPVRClientNextPVR::~cPVRClientNextPVR()
   delete m_timeshiftBuffer;
   delete m_recordingBuffer;
   delete m_realTimeBuffer;
+  m_epgOidLookup.clear();
+  m_hostFilenames.clear();
+  m_channelTypes.clear();
+  m_liveStreams.clear();
 }
 
 std::vector<std::string> cPVRClientNextPVR::split(const std::string& s, const std::string& delim, const bool keep_empty)
@@ -345,6 +349,10 @@ bool cPVRClientNextPVR::Connect()
                   m_iDefaultPrePadding = atoi(settingsDoc.RootElement()->FirstChildElement("PrePadding")->FirstChild()->Value());
                   m_iDefaultPostPadding = atoi( settingsDoc.RootElement()->FirstChildElement("PostPadding")->FirstChild()->Value());
                 }
+
+                m_showNew = false;
+                XMLUtils::GetBoolean(settingsDoc.RootElement(),"ShowNewInGuide",m_showNew);
+                XBMC->Log(LOG_NOTICE, "Show new %d",m_showNew);
 
                 if ( settingsDoc.RootElement()->FirstChildElement("RecordingDirectories") != NULL &&  settingsDoc.RootElement()->FirstChildElement("RecordingDirectories")->FirstChild() != NULL)
                 {
@@ -605,27 +613,22 @@ PVR_ERROR cPVRClientNextPVR::GetEpg(ADDON_HANDLE handle, int iChannelUid, time_t
         broadcast = {};
 
         string title;
+        title.clear();
         string description;
+        description.clear();
         string subtitle;
+        subtitle.clear();
         XMLUtils::GetString(pListingNode,"name",title);
-        if ( ! XMLUtils::GetString(pListingNode,"description",description))
-          description = "";
+        XMLUtils::GetString(pListingNode,"description",description);
 
         if (XMLUtils::GetString(pListingNode,"subtitle",subtitle))
         {
-          if (description == subtitle + ":")
-          {
-            description.clear();
-          }
-          else if (StringUtils::StartsWith(description,subtitle + ": "))
+          if (description != subtitle + ":" && StringUtils::StartsWith(description,subtitle + ": "))
           {
             description = description.substr(subtitle.length()+2);
           }
         }
-        else
-        {
-          subtitle.clear();
-        }
+
         broadcast.iYear = XmlGetInt(pListingNode,"year");;
 
         string startTime, endTime;
@@ -633,12 +636,17 @@ PVR_ERROR cPVRClientNextPVR::GetEpg(ADDON_HANDLE handle, int iChannelUid, time_t
         startTime.resize(10);
         XMLUtils::GetString(pListingNode,"end",endTime);
         endTime.resize(10);
-        broadcast.iUniqueBroadcastId = XmlGetInt(pListingNode,"id");
+
+        const std::string oidLookup(endTime + ":" + to_string(iChannelUid));
+
+        const int epgOid = XmlGetInt(pListingNode,"id");
+        m_epgOidLookup[oidLookup] = epgOid;
 
         broadcast.strTitle            = title.c_str();
         broadcast.strEpisodeName      = subtitle.c_str();
         broadcast.iUniqueChannelId    = iChannelUid;
         broadcast.startTime           = stol(startTime);
+        broadcast.iUniqueBroadcastId  = stoi(endTime);
         broadcast.endTime             = stol(endTime);
         broadcast.strPlotOutline      = NULL; //unused
         broadcast.strPlot             = description.c_str();
@@ -653,7 +661,7 @@ PVR_ERROR cPVRClientNextPVR::GetEpg(ADDON_HANDLE handle, int iChannelUid, time_t
         artworkPath[0] = '\0';
         if (g_bDownloadGuideArtwork)
         {
-          snprintf(artworkPath, sizeof(artworkPath), "http://%s:%d/service?method=channel.show.artwork&sid=%s&event_id=%d", g_szHostname.c_str(), g_iPort, m_sid, broadcast.iUniqueBroadcastId);
+          snprintf(artworkPath, sizeof(artworkPath), "http://%s:%d/service?method=channel.show.artwork&sid=%s&event_id=%d", g_szHostname.c_str(), g_iPort, m_sid, epgOid);
           broadcast.strIconPath         = artworkPath;
         }
         std::string sGenre;
@@ -697,7 +705,8 @@ PVR_ERROR cPVRClientNextPVR::GetEpg(ADDON_HANDLE handle, int iChannelUid, time_t
         {
           broadcast.iFlags = EPG_TAG_FLAG_IS_NEW;
         }
-        else
+
+        if (m_showNew)
         {
           bool firstrun;
           if (XMLUtils::GetBoolean(pListingNode,"firstrun",firstrun))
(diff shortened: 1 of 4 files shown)


emveepee:
@sub3 This is an interim fix in the hope that we can go back to using the backend event oid in Kodi.  It is also WIP but I wanted to test on several platforms and offer something to users who get stuck.  

I need to add the following

- figure out the best location to put an iterator to erase expired keys from the map
-  use standard c++ or Kodi addon libraries to get the epoch time for today at midnight utc to use for the the erase.  I don't really want to add the CDateTime modules just for this
- clear the map on exit

src/pvrclient-nextpvr.cpp
```diff
@@ -2111,7 +2104,10 @@ PVR_ERROR cPVRClientNextPVR::AddTimer(const PVR_TIMER &timerinfo)
   std::string encodedName = UriEncode(timerinfo.strTitle);
   std::string encodedKeyword = UriEncode(timerinfo.strEpgSearchString);
   std::string days = GetDayString(timerinfo.iWeekdays);
-  std:string directory = UriEncode( m_recordingDirectories[timerinfo.iRecordingGroup]);
+  std::string directory = UriEncode( m_recordingDirectories[timerinfo.iRecordingGroup]);
```
phunkyfish: These could be const if you want. Up to you

phunkyfish (approved):
Looks good, the minor is up to you ;)

emveepee:
> 
> 
> Looks good, the minor is up to you ;)

It will probably become a feature update before I'm done.  I would like to pull the backend database update time if @sub3 is able to persist it between server reboots and use that to  trigger EPG updates and help clean ups of most of the stale cache, time changes will linger until they age out.  

One other thing will be to change the lookup to use the end date.  I can't age out on start time since someone might record a 24 hour rally.

src/pvrclient-nextpvr.cpp
emveepee: Right, I did say I would change ares of code that I was touching so I will update those

phunkyfish:
Do you generally just request a review once you are ready?

emveepee:
> 
> 
> Do you generally just request a review once you are ready?

No I prefer a review as I submit but the approval may not be necessary if there are comments that indicate we go ahead.   With the monolithic nature of pvr-nextpvr.cpp it is hard to create multiple PR especially when I am touching the same small groups of code for this and for the API changes.  If comments come in after I request the tag that could cause extra work.

Normally I would not post WIP but this case is special because effectively the addon is broken, and I can't expect users to delete tables.   The state of the world also makes it wiser to be a little more open sharing code.

phunkyfish:
Cool. I can review as you submit so.

If something is WIP just say so and I can leave it till you are ready.

ksooo:
This is, like you said, a workaround. If Kodi or the addon gets restarted (due to addon update, disable/enable, uninstall/install), you might run into the "freezing" problem again.

If you look for a far more simple workaround, with does not suffer from the restart problem, just calculate the broadcast uid in the addon based on event's start date and time. Those UIDs only need to be unique within a channel, thus this simple algorithm works quite well. Only drawback is that if the start time of an event changes, its UID will change, but this is a rather rare edge case.

emveepee:
> 
> If you look for a far more simple workaround, with does not suffer from the restart problem, just calculate the broadcast uid in the addon based on event's start date and time. Those UIDs only need to be unique within a channel, thus this simple algorithm works quite well. Only drawback is that if the start time of an event changes, its UID will change, but this is a rather rare edge case.

That is what this does I just send the start time (will be end time) but I need to store the time:channel id as the key to the internal lookup to get the backend event id which I need to pass to the backend for EPG based recordings.

Where this method fails is as noted there is no why to tag a recording to and event since the time is not unique.

> 
> This is, like you said, a workaround. If Kodi or the addon gets restarted (due to addon update, disable/enable, uninstall/install), you might run into the "freezing" problem again.

Once the update is in place the date is in the iBroadcastUid column so it will just be the first Matrix to Matrix update that I am worried about and in pre alpha no big deal.  Worse case that I see is they corrupt the database not want to use two slow updates and they have to delete epg#.db

phunkyfish:
I have a question. Why not simply just set the broadcast id to the start or end time and leave the map out?

sub3:
> I have a question. Why not simply just set the broadcast id to the start or end time and leave the map out?

If he needs to call a NextPVR to schedule a recording etc, he has to pass it the event id. (so he has to have a way to get back from the channel:start-time to event id)

emveepee:
> 
> 
> I have a question. Why not simply just set the broadcast id to the start or end time and leave the map out?

I do, it is an int not string.  I still need the real backend UID to do an EPG based recording.

phunkyfish:
Ok, and maybe a silly question but why not use start or end time on the backend for the EPG IDs since they are generated daily?

(apologies it I have this incorrect)

emveepee:
> 
> 
> Ok, and maybe a silly question but why not use start or end time on the backend for the EPG IDs since they are generated daily?
> 
It would still need to be a key of time:channel  and some coding on the backend which is not open source. I also work on the Emby interface and it needs a unique key so it just moves the issue for me.   Also I think Kodi wants a unique key for iEpgEventId in PVR_RECORDINGS since iChannelUid is optional

A new API call could help but it would mean two methods in the addon code We support two backend versions and while v5 is much better than v4 it is missing cam support among other things so we can't force an update.

The issue ultimately will be fixing the reason for the workaround, the slow REPLACE INTO.  Even with this change a typical daily update is still noticeably slower than zapping the table first

phunkyfish:
No, I don't think Kodi needs a unique iEPGEventId for recordings (only unique per channel), but I'd say if it has that and the channel it can display the EPG info for the recording which can be more rich.

REPLACE_INTO is just INSERT or UPDATE right? I think if you split that into an insert and an update query it will take the same amount of time surely. I'm not sure this would be the culprit.

emveepee:
> 
> 
> No, I don't think Kodi needs a unique iEPGEventId for recordings (only unique per channel), but I'd say if it has that and the channel it can display the EPG info for the recording which can be more rich.

Your right https://github.com/xbmc/xbmc/blob/master/xbmc/pvr/recordings/PVRRecording.cpp#L480 but it doesn't handle the situation where the same show is being recording twice using different rules.   The code seems to already match on date and time anyway.

Thinking about it two recordings from the same show would break whether it was start time or broadcast id, so I can update the PR to use the time just to speed it up.

>
> REPLACE_INTO is just INSERT or UPDATE right? I think if you split that into an insert and an update query it will take the same amount of time surely. I'm not sure this would be the culprit.

That is for the issue thread but everything I have read says REPLACE INTO is slow and not indexed.  You could try to see the speed on one channel with sqlite 

    .open epg13.db
    update epgtags
    set  iBroadcastUid = random()
    where idEpg = 1;

Check the time for Kodi start and exit and check the time for the first channel.  Then run the same script go in clear the PVR database  and exit and check the same time.  If you think it is a cache issue exit Kodi, run delete epgtags and start and exit Kodi.

sub3:
> Ok, and maybe a silly question but why not use start or end time on the backend for the EPG IDs since they are generated daily?
> 
> (apologies it I have this incorrect)

Because NextPVR is a large and complicated application, with large existing api, already built around the existing id system, which has worked perfectly fine for the past 12 or so years. Itdoesn't make sense to change this in the back for just Kodi. It should probably be implemented in the addon.

phunkyfish:
> > Ok, and maybe a silly question but why not use start or end time on the backend for the EPG IDs since they are generated daily?
> > (apologies it I have this incorrect)
> 
> Because NextPVR is a large and complicated application, with large existing api, already built around the existing id system, which has worked perfectly fine for the past 12 or so years. I don't understand why I would need to change this for just Kodi.

You don't need to change it at all. @emveepee mentioned in a previous post that he was going to request some changes.

phunkyfish:
Thanks for walking me through the PR ;)

emveepee:
> 
> 
> Thanks for walking me through the PR ;)

NP.  Related to the PR, when I update it to make the changes and support PR17515 is it better  to push a new commit,  amend the previous one, or amend the changes and also push a new commit for PR17515?

phunkyfish:
As they are not related they should be separate commits. But I don’t yet know when #17515 will get merged as it’s a breaking change so maybe don’t wait for it.

emveepee:
@sub3 I ran into another slowdown when I enabled EPG artwork for testing since the artwork URLs needs the EPG event oid, and there is no callback to modify the URL before passing it to NextPVR.  Users can either have speed or disable EPG art for now

Also if any other authors follow this if you want to use the Recording time to link to the EPG time note that only works in NextPVR for pending and in-progress recordings.  Once a recording starts the start time is clock time which could include pre-padding or a late start.   When the recording is stopped the duration is actual time so that can also including post padding or an early stop.   This is another reason to use end time for any key.  For ready recordings the time will almost always be wrong.

AlwinEsch:
About them it need a rebase related to version increase, this should then become 4.3.5. Was a bit fast on my request, Sorry.

emveepee:
@sub3 the second commit passed my tests on v5 with Live, New Premiere and Finale  you might want to review the conditions for the significance field since I am not sure how you populate non SD data.  Also there will be no historical date for significance on recordings but the mapping between the timer and the EPG is working well and New and Live flags do show on pending and in-progress recordings.

@phunkyfish Great job.  The first update did slow down the EPG update too because of 100's of changes from IsNew.   The only minor comment I have is for me Premiere looks out of place with the choice to go with the accent.

src/pvrclient-nextpvr.cpp
```diff
@@ -1295,6 +1309,18 @@ bool cPVRClientNextPVR::UpdatePvrRecording(TiXmlElement* pRecordingNode, PVR_REC
     PVR_STRCPY(tag->strGenreDescription,genres.c_str());
   }
 
+  std::string significance;
+  significance.clear();
```
phunkyfish: This is not required. The string will initialise to empty.

src/pvrclient-nextpvr.cpp
```diff
@@ -701,18 +701,32 @@ PVR_ERROR cPVRClientNextPVR::GetEpg(ADDON_HANDLE handle, int iChannelUid, time_t
...
+        if (XMLUtils::GetBoolean(pListingNode,"firstrun",firstrun))
         {
-          broadcast.iFlags = EPG_TAG_FLAG_IS_NEW;
-        }
-
-        if (m_showNew)
-        {
-          bool firstrun;
-          if (XMLUtils::GetBoolean(pListingNode,"firstrun",firstrun))
+          if (firstrun == true)
```
phunkyfish: `if (firstrun)`

phunkyfish (changes requested):
How come no check for Live and New in `UpdatePvrRecording`? Just curious

phunkyfish:
You also have `New` code in the first commit. Was that intentional?

src/pvrclient-nextpvr.cpp
emveepee: I misinterpreted something that ksooo asked me to do on an old conversation I am glad I don't have to do that anymore

src/pvrclient-nextpvr.cpp
emveepee: If that is recommended style I will use it

phunkyfish: It really is up to you but generally it’s common practice.

emveepee:
> How come no check for Live and New in `UpdatePvrRecording`? Just curious

Several reasons

-  New is not in the same field and I didn't ask sub to provide it on recording.
- I think Live and New look wrong on old recordings.  Does it actually make a difference if a recorded show was recorded live or new.  In resolving conflicts NextPVR even when it is not new
- The timer EPG links provide this information on pending and in progress recordings where it does matter.

- Additional the logic is not the same for recordings I show the Premiere and Recording for reruns because @sub3 asked me to do it.  That makes sense because there is no S/E order in Kodi and date doesn't always match for shows that are new.     Flagging these on the would be confusing.

emveepee:
> You also have `New` code in the first commit. Was that intentional?

Yes I changed what you suggested, and  I mentioned I was going to change to end date and I released I was using the fake iBroadcastUid for a call to the backend

phunkyfish:
Apologies. I wasn’t clear. I mean the `m_showNew` check. Is that supposed to be in the first commit or the second one?

emveepee:
> Apologies. I wasn’t clear. I mean the `m_showNew` check. Is that supposed to be in the first commit or the second one?

Ah right lots of changes recently.  Sub3 and I discussed this offline and he explained that when the guide data isn't clear on whether a show is new or not, he will mark it new.  There is a server side which is read that controls whether not to show all shows are new or not.

phunkyfish:
> Ah right lots of changes recently. Sub3 and I discussed this offline and he explained that when the guide data isn't clear on whether a show is new or not, he will mark it new. There is a server side which is read that controls whether not to show all shows are new or not.

Ok, cool, makes sense. It still sounds like it belongs in the second commit though right?

src/pvrclient-nextpvr.cpp
emveepee: Ideally this variable should have a null choice but yes I can change that

emveepee:
> 
> Ok, cool, makes sense. It still sounds like it belongs in the second commit though right?

If you mean from a review perspective yes but in reality I guess there should have been a third commit because I needed it whether or not the commit for PR17515 was released.  It's tough when one small function has three enhancements coming in different directions.  When I look at the code the changes I made for the second commit were crap.

src/pvrclient-nextpvr.cpp
phunkyfish: This should still be removed.

emveepee: Thanks I forget about it being in two places and @sub3 hasn't even uploaded the new version of the backend that supports this in Recordings and will push it in the next commit.   That reminds me I do have to actually schedule a recording with premiere/finale once the new version is available to test this, it won't be on historical data.

emveepee: Changed

src/pvrclient-nextpvr.cpp
emveepee: Changed

src/pvrclient-nextpvr.cpp
```diff
@@ -605,40 +613,40 @@ PVR_ERROR cPVRClientNextPVR::GetEpg(ADDON_HANDLE handle, int iChannelUid, time_t
         broadcast = {};
 
         string title;
+        title.clear();
```
phunkyfish: You remove these again in the following commit. Why not just not add them in the first place?

src/pvrclient-nextpvr.cpp
```diff
@@ -605,40 +613,40 @@ PVR_ERROR cPVRClientNextPVR::GetEpg(ADDON_HANDLE handle, int iChannelUid, time_t
...
         string description;
+        description.clear();
```
phunkyfish: And here

src/pvrclient-nextpvr.cpp
```diff
@@ -605,40 +613,40 @@ PVR_ERROR cPVRClientNextPVR::GetEpg(ADDON_HANDLE handle, int iChannelUid, time_t
...
         string subtitle;
+        subtitle.clear();
```
phunkyfish: And here

src/pvrclient-nextpvr.h
```diff
@@ -201,6 +201,8 @@ class cPVRClientNextPVR : P8PLATFORM::CThread
   std::map<std::string, std::string> m_hostFilenames;
   std::map<int, bool> m_channelTypes;  // returns isRadio
   std::map<int, std::string> m_liveStreams;
+  std::map<std::string,int> m_epgOidLookup;
+  bool                    m_showNew;
```
phunkyfish, line 205: Any reason for the many spaces here?

src/pvrclient-nextpvr.h
```diff
@@ -201,6 +201,8 @@ class cPVRClientNextPVR : P8PLATFORM::CThread
   std::map<std::string, std::string> m_hostFilenames;
   std::map<int, bool> m_channelTypes;  // returns isRadio
   std::map<int, std::string> m_liveStreams;
+  std::map<std::string,int> m_epgOidLookup;
```
phunkyfish, line 204: Just a general comment but the use of spacing just after commas appears in consistent with the existing code (such as the line above this). It catches my eye each time I see your code ;)

phunkyfish (commented):
Just comments, let me know if you find them useful. (Or I can shut up!)

src/pvrclient-nextpvr.cpp
emveepee: I believe you are reading them in the wrong order,  when I look at the current source they are not there.   I added them in the first commit based on your issue that my data wasn't being cleared out causing some EPG update pointer issues  but then after the second commit you asked me to remove them.  I didn't want to undo everything to go back to commit one to remove them and then recreate commit one and then two.

src/pvrclient-nextpvr.h
emveepee: I don't like it either but the original code is like that and if you look at it in context it isn't as bad https://github.com/emveepee/pvr.nextpvr/blob/Matrix/src/pvrclient-nextpvr.h#L205 

I would like to change them all one day but many of the pvr addons do the member variables like that if you look at their header files.  That will be one commit.

src/pvrclient-nextpvr.h
emveepee: I use a mix of linux and windows building and testing this code, so perhaps the autospacing is different in vs code or visual stdio I will watch for it, this doesn't bother me but I will change it but I won't do it this time, because it involves changing commit 1.

phunkyfish (approved):
👍

PR #116, opened 2020-03-25
[Matrix] cleanups, addon.xml update, icon assets add and LICENSE add / update

AlwinEsch:


emveepee:
@AlwinEsch the Windows installer doesn't really provide any license information to agree to

image

AlwinEsch:
This view related to Kodi itself where it takes https://github.com/xbmc/xbmc/blob/master/LICENSE.md
You are completely right thats wrong to show there.

But for here is it a wrong place, as it not on addon itself.
You can create a issue here https://github.com/xbmc/xbmc/issues

emveepee:
Ok thanks I wasn't sure if it was something I had to change, I will do that.  Also I would like to update the README.md to to match the linux build instructions (kodi not xbmc and no build folder) and also link to master build doc.  Is that file managed by the addon?

AlwinEsch:
You mean the https://github.com/kodi-pvr/pvr.nextpvr/blob/Matrix/README.md? 

If yes bring a request to see exact what you mean.

If related to here:
```
git clone --branch master https://github.com/xbmc/xbmc.git
git clone https://github.com/kodi-pvr/pvr.nextpvr.git
cd pvr.nextpvr && mkdir build && cd build
cmake -DADDONS_TO_BUILD=pvr.nextpvr -DADDON_SRC_PREFIX=../.. -DCMAKE_BUILD_TYPE=Debug -DCMAKE_INSTALL_PREFIX=../../xbmc/addons -DPACKAGE_ZIP=1 ../../xbmc/cmake/addons
make
```

With `xbmc` as Kodi repo still with this name and also about compatibility stay like them.
There can be as alternate `git clone --branch master https://github.com/xbmc/xbmc kodi` where the name becomes changed to `kodi`.

About `build` folder there is primary to prevent problems if something in Code becomes changed and to prevent by `git add` the wrong parts.

But to improve the README whould be good with instructions or something else.

emveepee:
The linux instructions say  git clone https://github.com/xbmc/xbmc kodi  so I want to want to a trivial update for that. 

On https://github.com/xbmc/xbmc/blob/master/docs/README.Linux.md the linux binary build step for pvr.demo in 5.2 is quite different then the pvr.demo README.md  I was going to test to see how that keeps everything in sync with xbmc

AlwinEsch:
To use source folder is not good to use by them :smile:, should really build folder there too.

```
cd ~/src
git clone https://github.com/kodi-pvr/pvr.demo.git
cd ~/src/pvr.demo/
cmake -DCMAKE_INSTALL_PREFIX=/usr/local
make && make install
```

About how a addon can be build specially by Linux are much different ways :unamused: where it becomes harder to instruct about them.

This there is thought to a Linux distribution, where all depends must be inside system, about pvr.demo this is not much, but if you take others can it be harder.

By README.md here on addon is related to Kodi's distribution independent build, where it create also all required depends itself, without take from OS. That's also why his install folder goes to Kodi's own build folder (where a bit harder to test and use then on system installed).

I take normally it as this:
```sh
mkdir build && cd build && cmake \
  -DADDONS_TO_BUILD=$(basename $(dirname $(pwd))) \
  -DADDON_SRC_PREFIX=../.. \
  -DCMAKE_BUILD_TYPE=Debug \
  -DCMAKE_INSTALL_PREFIX=$HOME/.kodi/addons \
  -DPACKAGE_ZIP=1  \
  ../kodi/cmake/addons && make
```
where it becomes installed inside Kodi's home folder and can be used from a system Kodi or from Development Kodi (the `PACKAGE_ZIP` define to have `lib` and `share` in same folder).

I recommend the one given on the addon or on my example, provided it is not intended for distributions. Distributions must then match the respective system.

PR #121, opened 2020-04-14
Matrix 4.4 feature enhancements

emveepee:
This PR will include multiple commits for a new 4.4 release of the addon.
New Matrix settings, new startup logic,  new options & features implementing improvements on the backend.


pvr.nextpvr/resources/settings.xml
```diff
@@ -1,25 +1,272 @@
...
+					<control type="toggle"/>
+					<dependencies>
+						<dependency type="visible">
+							<condition operator="is" setting="legacy">false</condition>
+						</dependency>
+					</dependencies>
+				</setting>
+			</group>
+		</category>
+	</section>
```
AlwinEsch: Thanks a lot for update them, only can you remove the tabs and use only spaces for it :thinking:

It looks on my Linux with tabs then e.g. here:
Bildschirmfoto vom 2020-04-15 04-58-36

And text copy produce `<------><------><------><------><------><level>1</level>`.

Therefore only please, but not as a request from me :smile:.

About the release of this addon, **maybe** I can combine your request with https://github.com/kodi-pvr/pvr.nextpvr/pull/119 for the version.

emveepee: I have a few commits coming for this PR so don't bother combining it as I am making a feature release.  Thanks for the offer.   Sorry about the tabs I do have VS Code setup properly in Ubuntu but I bet editing it in Windows today changed things.

emveepee: I found out how this happened VS Code doesn't modify a file when you change tab characteristics, it is display so saving don't save.  You have to run a convert and then save.  Thanks again

phunkyfish: One of the things I did a while back with some other addons was run clang format, fixed up and committed folder by folder.

It’s not perfect so you need to check the changes. I also increased the line length from kodi’s default. It was time consuming but worth it in the end as the whole codebase was consistent. Also made reviewing much easier. Just a thought.

emveepee: The documentation says not to do this for existing files but I will consider it as a standalone PR.  I prefer doing this when I somehow break the huge pvrclient-nextpvr.cpp into sub cpp modules similar to how is was done here https://github.com/kodi-pvr/pvr.nextpvr/pull/89/commits/7e42c9a7b7a2c8800edd110e840ce0bc496c5b47 and I don't want to include that with a feature release.

Does .clang-format apply to .xml and .po as well?

phunkyfish: For kodi the documentation says that. Addons do not have to obey those rules. Generally it’s one dev maintaining each so up to them. Fine to leave to a refactoring PR.

Clang-format only applies to C/C++/Objective C.

emveepee:
I'd like to get this out before the next API change, it is pain to build new versions of kodi on windows and ubuntu to test and resolve.  Are there any additional comments or concerns?

I do have an unrelated minor bug fix (problem is in Leia too) but should I add it here or in a new PR?

phunkyfish:
I will review tomorrow, please add the bug too, ping me once it ready for review.

src/buffers/TranscodedBuffer.h
```diff
@@ -20,8 +20,7 @@ namespace timeshift {
   public:
     TranscodedBuffer() : DummyBuffer()
     {
-      m_profile << "&profile=" << g_iResolution << "p";
-      XBMC->Log(LOG_INFO, "TranscodedBuffer created! %s", m_profile.str().c_str());
+      XBMC->Log(LOG_NOTICE, "TranscodedBuffer created");
```
phunkyfish: LOG_NOTICE should no longer be used in PVR addons in matrix

src/client.cpp
```diff
@@ -23,30 +23,33 @@ using namespace ADDON;
...
 cPVRClientNextPVR     *g_client       = NULL;
 std::string            g_szUserPath   = "";
 std::string            g_szClientPath = "";
-bool             g_bUseTimeshift;  /* obsolete but settings.xml might have it */
 
 CHelper_libXBMC_addon *XBMC           = NULL;
 CHelper_libXBMC_pvr   *PVR            = NULL;
 bool                   g_bDownloadGuideArtwork = false;
 
+int g_backendVersion{ 0 };
```
phunkyfish: Just a comment but you could consider introducing a settings class in the future instead of using global variables.

src/client.cpp
```diff
@@ -295,21 +284,31 @@ ADDON_STATUS ADDON_SetSetting(const char *settingName, const void *settingValue)
...
   else if (str == "livestreamingmethod")
   {
-    eStreamingMethod  setting_livestreamingmethod = *(eStreamingMethod*) settingValue;
-    if (g_livestreamingmethod == ClientTimeshift)
+    if (g_client)
     {
-        if (setting_livestreamingmethod == RealTime)
+      if (g_client->GetBackendVersion() < 50000)
+      {
+        eStreamingMethod  setting_livestreamingmethod = *(eStreamingMethod*)settingValue;
```
phunkyfish: One space and not two and C++ style cast

src/client.cpp
```diff
@@ -324,26 +323,17 @@ ADDON_STATUS ADDON_SetSetting(const char *settingName, const void *settingValue)
   }
   else if (str == "reseticons")
   {
-    if ( g_eraseIcons != *(bool*)settingValue )
+    if (*(bool*)settingValue == true)
```
phunkyfish: C++ style cast

src/client.cpp
```diff
@@ -324,26 +323,17 @@ ADDON_STATUS ADDON_SetSetting(const char *settingName, const void *settingValue)
...
   }
-  else if (str == "resolution")
+  else
   {
-    if (g_iResolution != *(int*) settingValue)
-    {
-      g_iResolution = *(int*) settingValue;
-      return ADDON_STATUS_NEED_RESTART;
-    }
+    XBMC->Log(LOG_DEBUG, "Skipping setting %s",str.c_str());
```
phunkyfish: Space after `,` for consistency

src/client.h
```diff
@@ -52,18 +52,22 @@ extern std::string      g_host_mac;
 extern int              g_wol_timeout;
 extern bool             g_wol_enabled;
 extern bool             g_KodiLook;
-extern bool             g_eraseIcons;
-extern int16_t          g_timeShiftBufferSeconds;
+extern int g_ServerTimeOffset;
+
+extern int              g_timeShiftBufferSeconds;
```
phunkyfish: aligned or not aligned?

src/client.h
```diff
@@ -52,18 +52,22 @@ extern std::string      g_host_mac;
...
 extern eStreamingMethod g_livestreamingmethod;
 extern eNowPlaying g_NowPlaying;
-extern int              g_iResolution;
+
```
phunkyfish: Extra line?

src/pvrclient-nextpvr.cpp
```diff
@@ -251,126 +248,62 @@ bool cPVRClientNextPVR::Connect()
...
+                if (XMLUtils::GetInt(settingsDoc.RootElement(), "SlipSeconds", g_timeShiftBufferSeconds))
+                  XBMC->Log(LOG_NOTICE, "time shift buffer in seconds == %d\n", g_timeShiftBufferSeconds);
+
+                std::string serverMac;
+                if (XMLUtils::GetString(settingsDoc.RootElement(),"ServerMAC",serverMac))
                 {
+                  // only available from Windows backend
                   char rawMAC[13];
-                  PVR_STRCPY(rawMAC,settingsDoc.RootElement()->FirstChildElement("ServerMAC")->FirstChild()->Value());
+                  PVR_STRCPY(rawMAC,serverMac.c_str());
```
phunkyfish: Space after comma

src/pvrclient-nextpvr.cpp
```diff
@@ -416,6 +353,100 @@ void cPVRClientNextPVR::Disconnect()
...
+
+  bool liveStreams;
+  if (XBMC->GetSetting("uselivestreams", &liveStreams))
+    if (liveStreams)
+      LoadLiveStreams();
+
+  bool resetIcons;
+  if (XBMC->GetSetting("reseticons", &resetIcons))
+  if (resetIcons)
+      DeleteChannelIcons();
```
phunkyfish: Indentation. Maybe use brackets for the outer `if` so the scope if clear.

src/pvrclient-nextpvr.cpp
```diff
@@ -416,6 +353,100 @@ void cPVRClientNextPVR::Disconnect()
...
+  }
+  else
+  {
+    g_sendSidWithMetadata = true;
+  }
+
+  bool liveStreams;
+  if (XBMC->GetSetting("uselivestreams", &liveStreams))
+    if (liveStreams)
+      LoadLiveStreams();
```
phunkyfish: Maybe use brackets for the outer `if` so the scope if clear.

src/client.cpp
```diff
@@ -778,15 +779,18 @@ PVR_ERROR GetStreamReadChunkSize(int* chunksize)
   return PVR_ERROR_SERVER_ERROR;
 }
 
+PVR_ERROR SetRecordingPlayCount(const PVR_RECORDING& recording, int count)
+{
+  XBMC->Log(LOG_DEBUG, "Play count %s %d", recording.strTitle,count);
```
phunkyfish: space after comma

src/pvrclient-nextpvr.cpp
```diff
@@ -686,13 +686,22 @@ PVR_ERROR cPVRClientNextPVR::GetEpg(ADDON_HANDLE handle, int iChannelUid, time_t
...
+          else
+          {
+            if (g_sendSidWithMetadata)
+              snprintf(artworkPath, sizeof(artworkPath), "http://%s:%d/service?method=channel.show.artwork&sid=%s&name=%s", g_szHostname.c_str(), g_iPort, m_sid,UriEncode(title).c_str());
+            else
+              snprintf(artworkPath, sizeof(artworkPath), "http://%s:%d/service?method=channel.show.artwork&name=%s", g_szHostname.c_str(), g_iPort, UriEncode(title).c_str());
+            if (g_eventArtFormat == Portrait)
+              strcat(artworkPath, "&prefer=poster");
+          }
+          broadcast.strIconPath = artworkPath;
```
phunkyfish: Do you use StringUtils::Format else in the addon? Might be better to use that as you don't have to worry about running out of buffer space.

Just a suggestion mind you, it's up to you.

src/pvrclient-nextpvr.cpp
```diff
@@ -1317,10 +1326,23 @@ bool cPVRClientNextPVR::UpdatePvrRecording(TiXmlElement* pRecordingNode, PVR_REC
...
+    else
+    {
+      if (g_sendSidWithMetadata)
+        snprintf(artworkPath, sizeof(artworkPath), "http://%s:%d/service?method=channel.show.artwork&sid=%s&name=%s", g_szHostname.c_str(), g_iPort, m_sid, UriEncode(tag->strTitle).c_str());
+      else
+        snprintf(artworkPath, sizeof(artworkPath), "http://%s:%d/service?method=channel.show.artwork&name=%s", g_szHostname.c_str(), g_iPort, UriEncode(tag->strTitle).c_str());
+      PVR_STRCPY(tag->strFanartPath, artworkPath);
+      strcat(artworkPath, "&prefer=poster");
+      PVR_STRCPY(tag->strThumbnailPath, artworkPath);
+    }
```
phunkyfish: Maybe a future item to remove the horrible PVR_STRCPY macro usage.

src/pvrclient-nextpvr.cpp
```diff
@@ -636,7 +636,7 @@ PVR_ERROR cPVRClientNextPVR::GetEpg(ADDON_HANDLE handle, int iChannelUid, time_t
       XBMC->Log(LOG_DEBUG, "Skipping expired EPG data %d %ld %lld",iChannelUid,iStart, iEnd);
       return PVR_ERROR_INVALID_PARAMETERS;
   }
-  sprintf(request, "/service?method=channel.listings&channel_id=%d&start=%d&end=%d", iChannelUid, (int)iStart, (int)iEnd);
+  sprintf(request, "/service?method=channel.listings&channel_id=%d&start=%d&end=%d&genre=all", iChannelUid, (int)iStart, (int)iEnd);
```
phunkyfish: c++ style casts (consider use of StringUtils::Format in the future)

src/pvrclient-nextpvr.cpp
```diff
@@ -718,7 +718,22 @@ PVR_ERROR cPVRClientNextPVR::GetEpg(ADDON_HANDLE handle, int iChannelUid, time_t
           broadcast.iGenreType = XmlGetInt(pListingNode,"genre_type");
           broadcast.iGenreSubType = XmlGetInt(pListingNode,"genre_subtype");
         }
-
+        if (pListingNode->FirstChildElement("genres") != NULL)
+        {
+          std::string  allGenres;
```
phunkyfish: One space

src/pvrclient-nextpvr.cpp
```diff
@@ -1368,7 +1383,7 @@ bool cPVRClientNextPVR::UpdatePvrRecording(TiXmlElement* pRecordingNode, PVR_REC
   {
     std::string genres;
     TiXmlElement* pGenresNode = pRecordingNode->FirstChildElement("genres");
-    GetAdditiveString(pGenresNode,"genre",", ",genres,true);
+    GetAdditiveString(pGenresNode,"genre",EPG_STRING_TOKEN_SEPARATOR,genres,true);
```
phunkyfish: Spaces after commas

src/Settings.cpp
```diff
@@ -0,0 +1,339 @@
...
+      XMLUtils::GetInt(settingsDoc.RootElement(), "PostPadding", m_defaultPostPadding);
+
+      m_showNew = false;
+      XMLUtils::GetBoolean(settingsDoc.RootElement(),"ShowNewInGuide",m_showNew);
+
+      std::string recordingDirectories;
+      if (XMLUtils::GetString(settingsDoc.RootElement(),"RecordingDirectories",recordingDirectories))
+      {
+        m_recordingDirectories = StringUtils::Split(recordingDirectories,",",0);
+        /*
```
phunkyfish, line 145: Remove commented code or add a TODO:

src/Settings.cpp
```diff
@@ -0,0 +1,339 @@
...
+        }
+      }
+    }
+  }
+  return ADDON_STATUS_OK;
+}
+
+void Settings::SetVersionSpecificSettings()
+{
+
```
phunkyfish: Extra line

src/client.cpp
```diff
@@ -227,114 +149,23 @@ void ADDON_ReadSettings(void)
...
-  else if (str == "reseticons")
-  {
-    if (*(bool*)settingValue == true)
-    {
-      XBMC->Log(LOG_INFO, "Flagging icon reset");
-      return ADDON_STATUS_NEED_RESTART;
-    }
-  }
-  else
+  ADDON_STATUS  status = settings.SetValue(settingName, settingValue);
```
phunkyfish: 2 spaces

src/pvrclient-nextpvr.cpp
```diff
@@ -789,10 +798,17 @@ PVR_ERROR cPVRClientNextPVR::GetChannels(ADDON_HANDLE handle, bool bRadio)
   std::string      stream;
   LOG_API_CALL(__FUNCTION__);
 
-  m_channelTypes.clear();
+  std::map<int, std::pair<bool, bool>>::iterator  itr = m_channelDetails.begin();
+  while (itr != m_channelDetails.end()) {
```
phunkyfish: Bracket on next line

src/pvrclient-nextpvr.cpp
```diff
@@ -789,10 +798,17 @@ PVR_ERROR cPVRClientNextPVR::GetChannels(ADDON_HANDLE handle, bool bRadio)
...
+    if ( itr->second.second == (bRadio == true) )
```
phunkyfish: Spaces not needed at start/end of round brackets: `if (itr->second.second == (bRadio == true))`

src/pvrclient-nextpvr.cpp
```diff
@@ -1121,7 +1133,7 @@ PVR_ERROR cPVRClientNextPVR::GetRecordings(ADDON_HANDLE handle)
   return returnValue;
 }
 
-bool cPVRClientNextPVR::UpdatePvrRecording(TiXmlElement* pRecordingNode, PVR_RECORDING *tag)
+bool cPVRClientNextPVR::UpdatePvrRecording(TiXmlElement* pRecordingNode, PVR_RECORDING *tag, std::string title, bool flatten)
```
phunkyfish: `const std::string& title`?

src/pvrclient-nextpvr.h
```diff
@@ -124,7 +124,7 @@ class cPVRClientNextPVR : P8PLATFORM::CThread
   int GetRecordingLastPlayedPosition(const PVR_RECORDING &recording);
   PVR_ERROR GetRecordingEdl(const PVR_RECORDING& recording, PVR_EDL_ENTRY[], int *size);
   PVR_ERROR GetRecordingStreamProperties(const PVR_RECORDING*, PVR_NAMED_VALUE*, unsigned int*);
-  bool UpdatePvrRecording(TiXmlElement* pRecordingNode, PVR_RECORDING *tag);
+  bool UpdatePvrRecording(TiXmlElement* pRecordingNode, PVR_RECORDING *tag, std::string title, bool flatten);
```
phunkyfish: `const std::string& title`?

phunkyfish (changes requested):
Ignore my comment on a Settings class, I see you did that in a furthet commit ;)

Also for future work the the settings format also supports a help info string. Really useful for your users and can save you a lot on support burden.

src/buffers/TranscodedBuffer.h
emveepee: Fixed in a commit after 4.3.7

src/client.h
emveepee: Those were removed in additional commit.  However my philosophy is to only change alignment when I touch the code.

src/pvrclient-nextpvr.cpp
emveepee: It was changed in a newer commit with braces and indentation

src/pvrclient-nextpvr.cpp
emveepee: Makes sense.  Changed

src/pvrclient-nextpvr.cpp
emveepee: Yes I am migrating away from snprintf That strcat worries me now since I guess there is a small possibility that someone might use a ridiculous hosthame.   The current code would crash in that case too.  I will change this to  ::Format()

src/pvrclient-nextpvr.cpp
emveepee: That is definitely in may plans as I move from the pvr-vuplus style coding I started with Settings  I also will run clang-format on all the new modules to deal with the style issues.

src/Settings.cpp
emveepee: Actually that was the old code left in to remind myself to test that Split did work as expected and I forgot to remove it.  Done

src/Settings.cpp
emveepee: Done

src/client.cpp
emveepee: OK

src/pvrclient-nextpvr.cpp
emveepee: OK

src/pvrclient-nextpvr.h
emveepee: OK

src/pvrclient-nextpvr.cpp
emveepee: OK

src/pvrclient-nextpvr.cpp
emveepee: OK

src/pvrclient-nextpvr.cpp
emveepee: OK

src/pvrclient-nextpvr.cpp
emveepee: OK

src/pvrclient-nextpvr.cpp
emveepee: Changed the cast,  I plan a further review to change the DoRequest function to and have the first parameter take a string.

src/client.cpp
emveepee: OK

src/pvrclient-nextpvr.cpp
emveepee: Changed same issue in newer module

src/client.h
emveepee: Old code

src/client.cpp
emveepee: Old code

src/client.cpp
emveepee: Fixed with move to Settings.cpp templates

src/client.cpp
emveepee: Old Code

src/client.cpp
emveepee: Already done as noted

src/client.cpp
emveepee: OK

src/pvrclient-nextpvr.cpp
emveepee: OK

src/pvrclient-nextpvr.cpp
emveepee: OK

src/pvrclient-nextpvr.h
emveepee: OK

src/pvrclient-nextpvr.cpp
emveepee: OK

src/client.cpp
emveepee: Old code

src/client.cpp
emveepee: Old Code

src/client.cpp
emveepee: Old code

src/client.cpp
emveepee: Already done

emveepee:
@phunkyfish  thank you for the review I didn't merge it with code as it impacted too many individual commits.  I hope that is acceptable and makes it easier to review the resolved issues.

phunkyfish:
A fix up commit like you have done is ok. Give me some time to review.

src/pvrclient-nextpvr.cpp
```diff
@@ -514,7 +516,7 @@ PVR_ERROR cPVRClientNextPVR::GetEpg(ADDON_HANDLE handle, int iChannelUid, time_t
       XBMC->Log(LOG_DEBUG, "Skipping expired EPG data %d %ld %lld",iChannelUid,iStart, iEnd);
       return PVR_ERROR_INVALID_PARAMETERS;
   }
-  sprintf(request, "/service?method=channel.listings&channel_id=%d&start=%d&end=%d&genre=all", iChannelUid, (int)iStart, (int)iEnd);
+  sprintf(request, "/service?method=channel.listings&channel_id=%d&start=%ld&end=%ld&genre=all", iChannelUid, static_cast<long int> (iStart), static_cast<long int> (iEnd));
```
phunkyfish: For you cast just this: `static_cast<long int>(iStart)`. So no space between `>` and `(`.

Why long int instead of int?

phunkyfish (changes requested):
Just a minor, nearly there.

src/pvrclient-nextpvr.cpp
emveepee: OK I can fix the space.  I used long int because PVR defines the start and end time as time_t

phunkyfish: Then you are better off to use `%lld` and `long long`.

phunkyfish: That will be the same on all platforms.

emveepee: OK I made those changes, I don't expect the add-on or the backend will work past 2038 without change.  Thanks again I will force it through if there are not other comments.

phunkyfish: My point was that `int` and `long int`. are the same. Both guarantee at least 32 bits.

If you wanted to be correct which I assume you did `long long` is correct. Using `long int` makes no sense.

emveepee: Ok I will just use int like other pvr's thanks.

phunkyfish (approved):

emveepee:
I am running into crashing problems with the Code Review commit I think I have a bad pointer with the move to String::Format for artworkPath.

Also @phunkyfish  what is the purpose of GetRecordingSize() polling when no recording is in progress?  I have been simply returning PVR_ERROR_NOT_IMPLEMENTED and this wasn't a problem but I feel that Kodi might not be liking it when  pCapabilities->bSupportsRecordingSize is true.

phunkyfish:
Where exactly does it crash?

Have you run it through a debugger?

phunkyfish:
The GetRecordingSizeCalls are fine, that's normal.

phunkyfish:
Should `m_sid` be `m_sid.c_str()`?

emveepee:
> 
> 
> The GetRecordingSizeCalls are fine, that's normal.

Thanks,

> 
> 
> Should `m_sid` be `m_sid.c_str()`?

no m_sid  is char[]

> 
> Where exactly does it crash? Have you run it through a debugger?
> 

No one spot and I haven't duplicated it today so I maybe moving artworkPath https://github.com/kodi-pvr/pvr.nextpvr/pull/121/commits/d50b84cd55935adf299fb155f2978013ab518b9e#diff-d502357537ac071b04080ad67ea5625fR575 before the if was necessary

In the debugger it was in a map related function in xtree.cpp on Visual Studio so I need to check to see if my maps are working.   Since I can't duplicate it this morning it I am not sure

- randomly on loading,  
- on exit which might be the spdlog bug and usually 
- on idle when the only thing that is happening in the log is Updating recordings size

I do have regular crashes in Matrix on clear pvr database on what seems to be the same error which may be a bug related to the serious issue that needs addressing.  Before clearing the PVR manage, Kodi has to persist new data and that can can a long minutes.  I haven't got around to looking into this.  I know that my destructor is not being called after the clear which I think is an issue but I don't know how PVR notifies me of the restart.

I decided to investigated this and now I see the issue, it was a stupid typo forgot to reset the itr here
https://github.com/kodi-pvr/pvr.nextpvr/pull/121/commits/d50b84cd55935adf299fb155f2978013ab518b9e#diff-d502357537ac071b04080ad67ea5625fR804

I have made these two changes so I think the crashing issues were resolved

It would be nice if after a reset Kodi could notify the client a reset was happening.

phunkyfish:
Did you find the issue?

emveepee:
> 
> 
> Did you find the issue?

Yes thanks, moving the artworkPath out of an "if" to keep it from being de-referenced probably solved one issue and iterating properly helped another .  

Debugging this I noticed that Kodi doesn't reset the  client when a database reset occurs so the preexisting logic of keeping a global variable for GetNumChannels() was wrong. I will rework  https://github.com/kodi-pvr/pvr.nextpvr/pull/121/commits/7a6d1cdcf8cd368637b9b93d0682fb4550d18506 and resubmit it with the Code Review commit unless  you can share a way to modify a previous commit..

phunkyfish:
You can modify any previous commit by rebasing.

For instance if the commit you want to edit is 7 commits back then run `git rebase -i HEAD~7`. Then change the commit in question from `pick` to `edit`. Next make your changes, save them and then run `git add -u` and then `git commit --amend`. Finish it off with `git rebase --continue`.

In english, you are rewinding to the commit you want to edit, changing it, recommitting it and applying all the later commits on top of it. Just note that if any of changes cause conflicts in a later commit you will need to fix those as you go.

emveepee:
> You can modify any previous commit by rebasing.
> 
> For instance if the commit you want to edit is 7 commits back then run `git rebase -i HEAD~7`. Then change the commit in question from `pick` to `edit`. Next make your changes, save them and then run `git add -u` and then `git commit --amend`. Finish it off with `git rebase --continue`.
> 
> In english, you are rewinding to the commit you want to edit, changing it, recommitting it and applying all the later commits on top of it. Just note that if any of changes cause conflicts in a later commit you will need to fix those as you go.

Thanks that is so helpful.  It looks like the GetNumChannel commit is what I want, but the build system did't like it.  Missed one in a later commit which you probably suspected.  Cherry picking is not for the faint of heart and all good now.

emveepee:
@AlwinEsch I have to add XBMC->CreateDirectory()  because without settings.xml (which does work) I don't have the user folder which holds channel icons

However XBMC->CreateDirectory("special://userdata/addon_data/pvr.nextpvr/"); won't compile on Windows because  of the Windows define

#define CreateDirectory  CreateDirectoryA

If I #undef CreateDirectory this works  Is that how this is supposed to work?

@phunkyfish  if you have an alternate solution for creating the user folder let me know.

src/pvrclient-nextpvr.h
```diff
@@ -112,7 +124,7 @@ class cPVRClientNextPVR : P8PLATFORM::CThread
   int GetRecordingLastPlayedPosition(const PVR_RECORDING &recording);
   PVR_ERROR GetRecordingEdl(const PVR_RECORDING& recording, PVR_EDL_ENTRY[], int *size);
   PVR_ERROR GetRecordingStreamProperties(const PVR_RECORDING*, PVR_NAMED_VALUE*, unsigned int*);
-  bool UpdatePvrRecording(TiXmlElement* pRecordingNode, PVR_RECORDING *tag);
+  bool UpdatePvrRecording(TiXmlElement* pRecordingNode, PVR_RECORDING *tag, const std::string title, bool flatten);
```
phunkyfish: `const std::string& title`

src/pvrclient-nextpvr.cpp
```diff
@@ -1160,13 +1124,12 @@ PVR_ERROR cPVRClientNextPVR::GetRecordings(ADDON_HANDLE handle)
   return returnValue;
 }
 
-bool cPVRClientNextPVR::UpdatePvrRecording(TiXmlElement* pRecordingNode, PVR_RECORDING *tag)
+bool cPVRClientNextPVR::UpdatePvrRecording(TiXmlElement* pRecordingNode, PVR_RECORDING *tag, const std::string title, bool flatten)
```
phunkyfish: `const std:string& title`

phunkyfish (changes requested):

src/pvrclient-nextpvr.h
emveepee: OK I don't know why but I will change it.

src/pvrclient-nextpvr.cpp
emveepee: Likewise

phunkyfish:
You should always pass by reference if you don't need a copy of an object. Generally const objects should be passed by reference.

phunkyfish (approved):
Well done ;)

emveepee:
> 
> 
> Well done ;)

Thanks for all your input.  Now just need to know about CreateDirectory for the last commit on this.

> 
> 
> You should always pass by reference if you don't need a copy of an object. Generally const objects should be passed by reference.

Thanks for that tip.

phunkyfish:
If I remember correctly that I needed to do something like that in iptv simple. Maybe have a look there.

emveepee:
> 
> 
> If I remember correctly that I needed to do something like that in iptv simple. Maybe have a look there.

Basically I can put it it in modules where windows.h is not included.  I didn't notice anything specific in iptv simple that needs windows.h

Martin

phunkyfish:
I’m not sure now but I remember needing it. Maybe around Stats. Or something.

emveepee:
Hi I have sent 3 new commit's since the last review.  I read about menuhooks in a recent PR and they were new to me but after reading more they seemed like the best way to do the reset of a channels items.  I saw this in mythv and from there figured out how to present a list of backend IP's to the user on a new install.

phunkyfish:
Can I suggest maybe we move the newer commits out from this PR? A PR should be a completed collection of work you are interested in merging. It may never end if you keep adding commits ;)

emveepee:
> 
> 
> Can I suggest maybe we move the newer commits out from this PR? A PR should be a completed collection of work you are interested in merging. It may never end if you keep adding commits ;)

I understand what you are saying but Settings.cpp Settings.h and (effectively) Settings.xml are new files and I feel they should be be merged in final form which is what this PR is. There is no real value in a merge followed by immediate new PR and then the inevitable delay that sub3 imposes waiting for additional comments after yours.   During that delay there could be a translation PR which could only complicate things.  It would certainly have helped if I had known about menuhooks when I started this PR.

I could remove the Request class but I didn't really like this code https://github.com/kodi-pvr/pvr.nextpvr/pull/121/commits/a7f4ba62f0cba72b19323d16c62d335f41edf167#diff-34f48ad91ac6c202ac60b0348ae90e30R76 that is in this code and it makes for a cleaner implementation.

I am not expecting another change to this module and I am hoping that it can be merged and tagged fairly quickly.

phunkyfish:
If we are drawing a line after this then ok for this PR.

What I'm trying to get to is that in the future let's try and have completed PRs to review, as there is a duplicate of effort on my part and time is finite ;)

src/pvrclient-nextpvr.cpp
```diff
@@ -160,32 +160,72 @@ cPVRClientNextPVR::~cPVRClientNextPVR()
...
+  if (item.cat == PVR_MENUHOOK_CHANNEL && menuhook.iHookId == 1)
   {
-    result.push_back(s);
-    return result;
+    DeleteChannelIcon(item.data.channel.iUniqueId);
+    PVR->TriggerChannelUpdate();
   }
-  std::string::const_iterator substart = s.begin(), subend;
-  while (true)
+  else if (item.cat == PVR_MENUHOOK_RECORDING && menuhook.iHookId == 1)
```
phunkyfish: Use constants instead of numbers

src/pvrclient-nextpvr.cpp
```diff
@@ -160,32 +160,72 @@ cPVRClientNextPVR::~cPVRClientNextPVR()
...
-      result.push_back(temp);
-    }
-    if (subend == s.end())
-    {
-      break;
-    }
-    substart = subend + delim.size();
+    ForgetRecording(item.data.recording);
+  }
+  else if (item.cat == PVR_MENUHOOK_SETTING && menuhook.iHookId == 1)
```
phunkyfish: Use constants instead of numbers

src/pvrclient-nextpvr.cpp
```diff
@@ -160,32 +160,72 @@ cPVRClientNextPVR::~cPVRClientNextPVR()
...
+  {
+    DeleteChannelIcons();
+    PVR->TriggerChannelUpdate();
+  }
+  else if (item.cat == PVR_MENUHOOK_SETTING && menuhook.iHookId == 2)
```
phunkyfish: Use constants instead of numbers

src/pvrclient-nextpvr.cpp
```diff
@@ -160,32 +160,72 @@ cPVRClientNextPVR::~cPVRClientNextPVR()
...
+  {
+    PVR->TriggerChannelUpdate();
+  }
+  else if (item.cat == PVR_MENUHOOK_RECORDING && menuhook.iHookId == 3)
```
phunkyfish: Use constants instead of numbers

src/pvrclient-nextpvr.cpp
```diff
@@ -160,32 +160,72 @@ cPVRClientNextPVR::~cPVRClientNextPVR()
...
+  }
+  return PVR_ERROR_NO_ERROR;
+}
+
+void cPVRClientNextPVR::ConfigureMenuhook()
+{
+  PVR_MENUHOOK menuHook;
+  menuHook = { 0 };
+  menuHook.category = PVR_MENUHOOK_CHANNEL;
+  menuHook.iHookId = 1;
```
phunkyfish: Use constants instead of numbers

src/pvrclient-nextpvr.cpp
```diff
@@ -160,32 +160,72 @@ cPVRClientNextPVR::~cPVRClientNextPVR()
...
+  menuHook.iLocalizedStringId = 30183;
+  PVR->AddMenuHook(&menuHook);
+
+  menuHook = { 0 };
+  menuHook.category = PVR_MENUHOOK_SETTING;
+  menuHook.iHookId = 1;
```
phunkyfish: Use constants instead of numbers

src/pvrclient-nextpvr.cpp
```diff
@@ -160,32 +160,72 @@ cPVRClientNextPVR::~cPVRClientNextPVR()
...
+  menuHook.iLocalizedStringId = 30170;
+  PVR->AddMenuHook(&menuHook);
+
+  menuHook = { 0 };
+  menuHook.category = PVR_MENUHOOK_SETTING;
+  menuHook.iHookId = 2;
```
phunkyfish: Use constants instead of numbers

src/pvrclient-nextpvr.cpp
```diff
@@ -160,32 +160,72 @@ cPVRClientNextPVR::~cPVRClientNextPVR()
...
+  menuHook.iLocalizedStringId = 30185;
+  PVR->AddMenuHook(&menuHook);
+
+  menuHook = { 0 };
+  menuHook.category = PVR_MENUHOOK_SETTING;
+  menuHook.iHookId = 3;
```
phunkyfish: Use constants instead of numbers

src/pvrclient-nextpvr.cpp
```diff
@@ -160,32 +160,72 @@ cPVRClientNextPVR::~cPVRClientNextPVR()
...
+  menuHook.iLocalizedStringId = 30186;
+  PVR->AddMenuHook(&menuHook);
+
+  if (m_settings.m_backendVersion >= 50000)
+  {
+    menuHook = { 0 };
+    menuHook.category = PVR_MENUHOOK_RECORDING;
+    menuHook.iHookId = 1;
```
phunkyfish: Use constants instead of numbers

src/BackendRequest.cpp
```diff
@@ -99,9 +100,67 @@ namespace NextPVR
...
+        XBMC->QueueNotification(QUEUE_INFO, XBMC->GetLocalizedString(30182), m_settings.m_hostname.c_str(), m_settings.m_port);
+        char *settings = XBMC->TranslateSpecialProtocol("special://profile/addon_data/pvr.nextpvr/settings.xml");
+        // create settings.xml with host and port
+        void* settingsFile = XBMC->OpenFileForWrite(settings,false);
+        const char tmp[] = "<settings version=\"2\">\n</settings>\n";
+        XBMC->WriteFile(settingsFile,tmp, strlen(tmp));
+        XBMC->CloseFile(settingsFile);
+        m_settings.SaveSettings("host",m_settings.m_hostname);
+        m_settings.SaveSettings("port",std::to_string(m_settings.m_port));
+        XBMC->FreeString(settings);
```
phunkyfish, line 154: Can you explain this, it seem really hacky and I don't think should be allowed.

src/BackendRequest.cpp
```diff
@@ -136,10 +195,7 @@ namespace NextPVR
             if (parseResponse.size() >= 3)
             {
               XBMC->Log(LOG_INFO, "Broadcast received %s %s", parseResponse[0].c_str(), parseResponse[1].c_str());
-              if (foundAddress.empty())
-              {
-                foundAddress = parseResponse;
-              }
+              foundAddress.push_back (parseResponse);
```
phunkyfish: Remove space

emveepee:
@phunkyfish I understand your concern about time, but there does seem to be a long delay at the review stage that leads to these.  When I first released this there was a week of no comments when I hoped for a merge  and then a few rounds like this with days without comment.  Half of the 21 days for this PR are simply waiting for @sub3 and you to review.   I would have hoped that this could have been merged and tagged by now so alpha test could be done in earnest.

I already have started on my next big change  which I have mentioned (separating timers, recordings and channels into classes with many clean ups ) but with everything jamming up at the review phase the commits might appear to stack up.

phunkyfish:
This is the reality though of this open source project (and many others). I'm a volunteer and do this in my spare time. That means PRs can take time to get through. 

Just finishing the next review of this PR.

phunkyfish (approved):
Manually editing settings.xml is a big no no. This change on settings should be deferred until we get the C++ API and this is possible via the API.

Other than that just minors.

src/BackendRequest.cpp
emveepee: > 
> 
> Can you explain this, it seem really hacky and I don't think should be allowed.

I am creating the default settings.xml with discovered IP address and port.  If I don't do this when the user goes to edit the settings it will be 127.0.0.1 since the read will use the settings.xml default.   It is only created when the folder doesn't existing so Kodi hasn't had a chance to read it.  After this is loaded the API calls access the file ok.

I have been updating the settings.xml with the MAC address for WOL for over one year with now issue.

phunkyfish: What happens if a user resets settings to default?

emveepee: This is the user settings.xml so default="true" won't be set by settings.  The discovered IP is never the default 127.0.0.1 so it will act like a user entered value.  If the user does click defaults it will get default from the master.

Here is it in action https://www.youtube.com/watch?v=XTkQLdKGF4M

phunkyfish: It's still a hack though. Would it not be better to just discover it each time?

emveepee: Discoverying every time would be a hack and would confuse a user with 127.0.0.1 in settings when they want to customize turn on time shift etc.  It is also leads to potential issues if a different backend was discovered without proper updating. Finally discovery is slower then a socket open.

While you consider it a hack, updating the ADDON API to support configurable default settings for this is not a small undertaking.

phunkyfish: Agreed, it’s already in the C++ API, so the ability will come
When that does.

I don’t there there is any workaround for this currently outside of what you have done (I still don’t have to like it 😉).

phunkyfish:
I was just chatting to @AlwinEsch and it looks like the C++ migration will happen in about 2 weeks. It’s pretty extensive so we will need to co-ordinate PRs around that time.

Do you think your next PR will be in by then?

src/BackendRequest.cpp
emveepee: Ok I will keep this in and gladly migrate it when the new API comes out.

src/pvrclient-nextpvr.cpp
emveepee: Done

src/pvrclient-nextpvr.cpp
emveepee: Done

src/pvrclient-nextpvr.cpp
emveepee: Done

src/pvrclient-nextpvr.cpp
emveepee: Done

src/pvrclient-nextpvr.cpp
emveepee: Done

src/pvrclient-nextpvr.cpp
emveepee: Done

src/pvrclient-nextpvr.cpp
emveepee: Done

src/pvrclient-nextpvr.cpp
emveepee: Done

src/BackendRequest.cpp
emveepee: Done

sub3:
> @phunkyfish I understand your concern about time, but there does seem to be a long delay at the review stage that leads to these. When I first released this there was a week of no comments when I hoped for a merge and then a few rounds like this with days without comment. Half of the 21 days for this PR are simply waiting for @sub3 and you to review. I would have hoped that this could have been merged and tagged by now so alpha test could be done in earnest.
> 

I usually hold off merging it for a bit to give others a chance to comment, but I'm always happy to be rushed, particularly for smaller less controversial changes. Just post a 'any comments?' reminder, and give me a bump if you'd like to get things moving.

emveepee:
> I was just chatting to @AlwinEsch and it looks like the C++ migration will happen in about 2 weeks. It’s pretty extensive so we will need to co-ordinate PRs around that time.
> 
> Do you think your next PR will be in by then?

I have TImers done and that was the biggest one, I expect after this PR  is merged and tagged I can focus on it and get it submitted in a few days.  I am sending all the modules through clang-format so that should make your  review easier.  Typically it is changing native tinyxml to XMLUtils,  getting rid of PVR_STRCPY where I can and changing some things to C++.  I might do some snprintf changes too when it makes sense.

phunkyfish:
Cool, sounds like a plan. I will let Alwin know not to do the nextpvr PR till closer to the merge date.

emveepee:
> 
> 
> I usually hold off merging it for a bit to give others a chance to comment, but I'm always happy to be rushed, particularly for smaller less controversial changes. Just post a 'any comments?' reminder, and give me a bump if you'd like to get things moving.

I am not sure if simply parking a PR for days after @phunkyfish approved it serves much purpose unless you actually call on other reviewers.   Basically I am just waiting for the merge and tag.

PR #125, opened 2020-05-14
Style edit and performance change

emveepee:
New .cpp files were crlf, changed to lf.   Change GetNumChannels() to use variables.  Kodi polls GetNumChannels() constantly when the Recordings menu is open.


phunkyfish:
Could you split the functional change and the refactoring into separate commits? There is no way to see what you have changed.

It's a good rule of thumb for any future work.

emveepee:
> Could you split the functional change and the refactoring into separate commits? There is no way to see what you have changed.
> 
> It's a good rule of thumb for any future work.

Ok hope this is easier.  I used dos2unix first then applied changes.   Not sure why Kodi  so aggressive in Recordings with CPVRClients::GetBackendProperties() polling so often.  I already don't bother getting drive space because of this.

phunkyfish (approved):
Nice

emveepee:
@sub3  when you have a chance could you merge this please.  I don't want to wait to bring the pause change and would like to get started with the changes for PR 16485

PR #126, opened 2020-05-14
3.3.20 Avoid seek after pause

emveepee:
No need to seek when there is buffer available

### DIFF
--- a/src/buffers/ClientTimeshift.cpp
+++ b/src/buffers/ClientTimeshift.cpp
@@ -115,6 +115,24 @@ void ClientTimeShift::Close()
   m_lastClose = time(nullptr);
 }
 
+void ClientTimeShift::Resume()
+{
+  ClientTimeShift::GetStreamInfo();
+  if (m_stream_duration > g_timeShiftBufferSeconds)
+  {
+    int64_t startSlipBuffer = m_stream_length - (g_timeShiftBufferSeconds * m_stream_length/m_stream_duration);
+    XBMC->Log(LOG_DEBUG, "%s:%d: %lld %lld %lld", __FUNCTION__, __LINE__, startSlipBuffer, m_streamPosition, m_stream_length.load());
+    if (m_streamPosition < startSlipBuffer)
+    {
+      Seek(m_streamPosition, 0);
+    }
+  }
+  else
+  {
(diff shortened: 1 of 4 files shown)


src/buffers/ClientTimeshift.cpp
```diff
@@ -115,6 +115,24 @@ void ClientTimeShift::Close()
...
+void ClientTimeShift::Resume()
+{
+  ClientTimeShift::GetStreamInfo();
+  if (m_stream_duration > g_timeShiftBufferSeconds)
+  {
+    int64_t startSlipBuffer = m_stream_length - (g_timeShiftBufferSeconds * m_stream_length/m_stream_duration);
+    XBMC->Log(LOG_DEBUG, "%s:%d: %lld %lld %lld", __FUNCTION__, __LINE__, startSlipBuffer, m_streamPosition, m_stream_length.load());
+    if (m_streamPosition < startSlipBuffer)
+    {
+      Seek(m_streamPosition,0);
```
phunkyfish: Space after comma

phunkyfish:
Other than the one minor just name the commit similar to the PR and it’s good.

src/buffers/ClientTimeshift.cpp
emveepee: Fixed

emveepee:
> 
> 
> Other than the one minor just name the commit similar to the PR and it’s good.

Ok thanks as always.  @sub3  if user feedback is positive this can be merged and tagged.

sub3:
Yep. First bit of feedback was positive, which is a good sign.

emveepee:
@sub3 since 18.7 is out I hope you don't hold off too much longer if you want this in any non binary distro like LibreElec.  There devs don't ask if we have versions pending.

sub3:
Can I safely ignore the jenkins 'this commit cannot be built'? If so, I'll merge it now.

emveepee:
The code seemed fine but since it is reasonable to see CI for macos I made a style change by removing a void in (void) in a header function definition to match the cpp 

https://github.com/kodi-pvr/pvr.nextpvr/pull/126/commits/0a8ed92a4a6efc04f0c980eec00cb87fd283be74#diff-e9db99c7c3cecea4006d71716d9397d6L94

and it is green now.

sub3:
Yep - I noticed that. I've merged it.

PR #127, opened 2020-05-16
[Matrix] API change update

AlwinEsch:
This changes are related to API change with xbmc/xbmc#17813

Need to bring in after them!

### DIFF
--- a/pvr.nextpvr/addon.xml.in
+++ b/pvr.nextpvr/addon.xml.in
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <addon
   id="pvr.nextpvr"
-  version="4.4.0"
+  version="5.0.0"
   name="NextPVR PVR Client"
   provider-name="Graeme Blackley">
   <requires>@ADDON_DEPENDS@</requires>
--- a/pvr.nextpvr/changelog.txt
+++ b/pvr.nextpvr/changelog.txt
@@ -1,3 +1,7 @@
+v5.0.0
+- Update PVR API 6.5.1
+- Update Global API 1.2.0
+
 v4.4.0
 - Move to Matrix settings.xml format enable new Kodi and backend options
 - Major restructuring of internal settings variables and new Settings class
--- a/src/client.cpp
+++ b/src/client.cpp
@@ -39,7 +39,7 @@ void ADDON_ReadSettings();
 // Called after loading of the dll, all steps to become Client functional
 // must be performed here.
 //-----------------------------------------------------------------------------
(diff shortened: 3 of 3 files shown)


emveepee:
@AlwinEsch why did you do this at this time?  Although I am grateful for what you do this doesn't help with the port for PR1648 that @phunkyfish  explained I was working on.  Hopefully it will fix a problem I am having with CreateDirectory() with PR1648t though.

AlwinEsch:
There was a part on filesystems parts needed to change (by request for @phunkyfish), where already has needed a update of all addons. For this reason are also few other, not so critical parts added, like the ADDON_LOG_... and the ADDON_Create to cleanup parts.

To split them everything in own requests makes it dramatic to big, if always every addon need update.

Also is this thought to have before PVR API change, as this not critical, but affect all addons. If now together, with the very critical PVR C++ rework becomes to much as also the eyes must on every other addon types then.

What you mean where with **PR1648**? **EDIT:** Do you mean https://github.com/xbmc/xbmc/pull/16485, I think yes :smile: 

**EDIT:** 
Related to PVR is nothing on **old** VFS changed, only need update as it also with his other headers affected to them on `versions.h`.

AlwinEsch:
This request have nothing todo with the PVR rework on Kodi. That's more about `ADDON_GLOBAL_VERSION_MAIN_MIN` where related on `AddonBase.h`, but also to `libXBMC_addon.h` and `xbmc_addon_dll.h` and thats why requests on PVR addons now.

phunkyfish:
@emveepee this change is required to go in prior to the C++ PVR work.

I thought I had mentioned that Alwin would be doing this before the other could be done.

phunkyfish:
Apologies I should have been clearer here: https://github.com/kodi-pvr/pvr.nextpvr/pull/124#issuecomment-628260645

You should be able to rebase after this PR goes in without it being too much work I hope.

AlwinEsch:
I can look to change that the log calls of PVR addons not affected.

emveepee:
It is really my mistake I didn't think the add on was going to be touched and all those ADDON_LOG_ message changes seem rather frivolous and extra work because XBMC->Log is being changed shortly anyway and I have changed them all so I expect this could make a rebase difficult

Edit @AlwinEsch  I just saw your post if you don't change the log message that would certainly be helpful.

AlwinEsch:
Is now a bit smaller :grinning: :smirk:

emveepee:
> 
> 
> Is now a bit smaller 😀 😏

Thanks so much, that is perfect.   I am hoping that when 6.5.1 is released, I will have the addon ready for C++

PR #128, opened 2020-05-18
C++ interface change WIP

emveepee:
Getting ready for API 7.0.0 there are a few issues before the submit

Needs kodi with PR 16485 to test


emveepee:
There are a few issues with this code it is still WIP   @phunkyfish  and @AlwinEsch I would be grateful for your comments.

The main issue is I cannot play a file.  This OpenFile does not open https://github.com/kodi-pvr/pvr.nextpvr/blob/07a8fb0143d3b30aa9be87b9883b74614927ec42/src/buffers/Buffer.cpp#L42 as it does in the C interface 

The log https://paste.kodi.tv/efeqiqakaq.kodi @ 2020-05-18 15:22:16.444 T:14500 shows the valid return from the server with an immediate close.  I have tried returning true, using CURLCreate,  and using the raw url but no changes.  

The other main new issue is in the loop that starts here https://github.com/kodi-pvr/pvr.nextpvr/blob/07a8fb0143d3b30aa9be87b9883b74614927ec42/src/BackendRequest.cpp#L36

I need to repack the buffer with 0's after each read.  This was not necessary previously

phunkyfish:
Stream.Read() will return the number of bytes read. You should only add that many bytes to the buffer right?

phunkyfish:
For m_inputHandle you create a locally scoped file the line above. That’s why it doesn’t work.

emveepee:
> 
> 
> Stream.Read() will return the number of bytes read. You should only add that many bytes to the buffer right?

I figured that is likely the problem, but with the C interface this wasn't necessary.
> 
> 
> For m_inputHandle you create a locally scoped file the line above. That’s why it doesn’t work.

Facepalm, removing that line works  since I have it in the base class too.  I probably was trying to figure out how to deal with the online documentation and the brackets.

/* Create the needed file handle class */
kodi::vfs::CFile myFile();

Can you think of a way to code this so the variable is defined in the base class but not created until when the file is opened is it really open created with OpenFile()

phunkyfish:
Just use a pointer but I don’t see why does it matter when it’s created?

src/buffers/Buffer.cpp
```diff
@@ -37,12 +38,12 @@ bool Buffer::Open(const std::string inputUrl, int optFlag)
     {
       ss << inputUrl;
     }
-    m_inputHandle = XBMC->OpenFile(ss.str().c_str(), optFlag );
+    kodi::vfs::CFile m_inputHandle;
```
AlwinEsch: With your file open, as you added `kodi::vfs::CFile` is it away after end of Open() call, This need to place inside a header.

AlwinEsch:
About file them can also hold without problems as `kodi::vfs::CFile m_inputHandle;` not direct need as pointer.

Them handle the "C" way, where was before inside addon. As long no Open was called the "C" pointer to Kodi stays there inside as `nullptr`. With the open becomes this then set in background and under following conditions removed:
- Addon calls the `Close()`
- Addon makes a new `Open()` then the before opened becomes automatic closed
- The related `kodi::vfs::CFile` class becomes destructed, the opened file becomes automatic closed.

src/buffers/Buffer.cpp
emveepee: Thanks @phunkyfish lead me to this too I had two reference to the variable.

emveepee:
> 
> 
> About file them can also hold without problems as `kodi::vfs::CFile m_inputHandle;` not direct need as pointer.
> 
> Them handle the "C" way, where was before inside addon. As long no Open was called the "C" pointer to Kodi stays there inside as `nullptr`. With the open becomes this then set in background and under following conditions removed:
> 
>     * Addon calls the `Close()`
> 
>     * Addon makes a new `Open()` then the before opened becomes automatic closed
> 
>     * The related `kodi::vfs::CFile` class becomes destructed, the opened file becomes automatic closed.

What about  when the addon closes and then tries to reopen the same file.  Here I close the file
https://github.com/kodi-pvr/pvr.nextpvr/blob/712b654a71ab0b3ed248bdc7b062ced10b63b51e/src/buffers/ClientTimeshift.cpp#L112 which does a m_inputHandle.Close() but when I reopen the file here https://github.com/kodi-pvr/pvr.nextpvr/blob/712b654a71ab0b3ed248bdc7b062ced10b63b51e/src/buffers/ClientTimeshift.cpp#L130   I am getting 

Exception thrown: read access violation.
this->**m_pFile** was 0xFFFFFFFFFFFFFFFF.

When the second OpenFile internally call it's own  Close().    Is it possible I have an old copy of the PR?  Even if I take out my Close() I get the same error.

AlwinEsch:
Can you make a backtrace (e.g. with `gdb`) to see which place exact makes this.

With `0xFFFFFFFFFFFFFFFF` looks like a `-1`was somewhere set, but should not done on pointer value :thinking:.

emveepee:
> 
> 
> Can you make a backtrace (e.g. with `gdb`) to see which place exact makes this.
> 
> With `0xFFFFFFFFFFFFFFFF` looks like a `-1`was somewhere set, but should not done on pointer value 🤔.

I figured it out, when I noticed it was File.cpp Close() and not Filesytem.h Close()   I mised part of the conversion I need to make this https://github.com/kodi-pvr/pvr.nextpvr/blob/712b654a71ab0b3ed248bdc7b062ced10b63b51e/src/buffers/Buffer.h#L192

void CloseHandle(kodi::vfs::CFile& handle);

Thanks for help.

emveepee:
@phunkyfish , @AlwinEsch @sub3   I am ready to re-push this for review.   Do you want me to simplify the for you,  perhaps by breaking it into multiple commits, addon structure change, log changes,  pvr api, vfs api, other api related changes.  There shouldn't be any functionality changes in this PR.

src/Settings.cpp
```diff
@@ -116,14 +111,14 @@ ADDON_STATUS Settings::ReadBackendSettings()
...
 
         // is the server new enough
         if (m_backendVersion < 40204)
         {
-          XBMC->Log(LOG_ERROR, "Your NextPVR version '%d' is too old. Please upgrade to '%s' or higher!", m_backendVersion, NEXTPVRC_MIN_VERSION_STRING);
-          XBMC->QueueNotification(QUEUE_ERROR, XBMC->GetLocalizedString(30050));
-          XBMC->QueueNotification(QUEUE_ERROR, XBMC->GetLocalizedString(30051), NEXTPVRC_MIN_VERSION_STRING);
+          kodi::Log(ADDON_LOG_ERROR, "Your NextPVR version '%d' is too old. Please upgrade to '%s' or higher!", m_backendVersion, NEXTPVRC_MIN_VERSION_STRING);
+          kodi::QueueNotification(QUEUE_ERROR, "Change - Need Header",  kodi::GetLocalizedString(30050));
+          //kodi::QueueNotification(QUEUE_ERROR, kodi::GetLocalizedString(30051), NEXTPVRC_MIN_VERSION_STRING);
```
AlwinEsch: About things like this gives it a QueueFormatedNotification, see http://alwinesch.github.io/group__cpp__kodi.html#ga39c65205ea763e00c402cb0f1f50b289

**Edit:** Maybe I add a function within kodi::tools part, to make a `kodi::tools::StringUtils::Format(...)`

phunkyfish:
> @phunkyfish , @AlwinEsch @sub3 I am ready to re-push this for review. Do you want me to simplify the for you, perhaps by breaking it into multiple commits, addon structure change, log changes, pvr api, vfs api, other api related changes. There shouldn't be any functionality changes in this PR.

A single commit should be ok.

src/Settings.cpp
emveepee: It turns out this is how I resolved it, I didn't know about QueueFormattedNotification.  

```
kodi::QueueNotification(QUEUE_ERROR, kodi::GetLocalizedString(30050), StringUtils::Format(kodi::GetLocalizedString(30051).c_str(), NEXTPVRC_MIN_VERSION_STRING));

**Comment by emveepee** [comment]
@AlwinEsch In doing my testing I noticed on the recording list there is a funny display from the Variables.xml variable  RecordingDateSizeLabel even in pvr.demo.  Is that fixed in a newer build of the PR?

**Comment by phunkyfish** [comment]
Please give a screenshot or example

**Comment by emveepee** [comment]
> 
> 
> Please give a screenshot or example

It  come from the ListItem.Size, (,) on 501 in Variables.xml This is pvr.demo

image

**Comment by AlwinEsch** [comment]
I look tommorow about and compare the old with new, possible that there still some errors in my new interface.

**Comment by emveepee** [comment]
> 
> 
> I look tommorow about and compare the old with new, possible that there still some errors in my new interface.

Thanks I am not sure if it is the new API, since my test build is about a week old, waiting for the rebase.   

I am also not get any requests to SetRecordingLastPlayedPosition() so not sure if that is hooked up or not.  I do report capabilities.SetSupportsLastPlayedPosition(true); and I see it getting sent back into Kodi.

**Comment by emveepee** [comment]
@AlwinEsch regarding the problem with the playback position, what is happening is the bookmark is always zero at exit.  NextPVR doesn't support GetLastPlayedPosition() on active files would that be a problem in the new code?  Right now it is just echoing back the old database values.  I did try not overriding GetPlayedPosition() but  that doesn't change the fact that it always exits here

https://github.com/xbmc/xbmc/blob/master/xbmc/Application.cpp#L3135

**Comment by emveepee** [comment]
@AlwinEsch  I now have run into a third issue in testing.  EDL skipping is working but I don't get the nice Matrix comskip markers showing on the timeline.    Is something required to trigger that?

@phunkyfish and @sub3  feel free to start review.  I will rebase against 5.0.1 when @sub3 merges and tags it and rebuild against Kodi master when PR 16485 is able to be built against master.

**Comment by emveepee** [comment]
I bumped everything to the current version of PR16485 and baseline kodi and still see the same 3 issues.

**Comment by AlwinEsch** [comment]
Have installed the `nextpvr` backend and runtime test (with few compile fixes) was OK.

About the EDL flags how they become from backend created? 
Has started recordings with it, but seems not given.

My patch about fix:
[0001-compile-fixes.patch.txt](https://github.com/kodi-pvr/pvr.nextpvr/files/4669658/0001-compile-fixes.patch.txt)

**Comment by emveepee** [comment]
> 
> 
> Have installed the `nextpvr` backend and runtime test (with few compile fixes) was OK.

> 
> My patch about fix:
> [0001-compile-fixes.patch.txt](https://github.com/kodi-pvr/pvr.nextpvr/files/4669658/0001-compile-fixes.patch.txt)

You must have missed my commit this morning since I updated client to handle those compile errors.  I will update the READ_NO_CACHE globally eventually.

> 
> About the EDL flags how they become from backend created?
> Has started recordings with it, but seems not given.

The edl files are created on the server by comskip.  You can make one manually as filename.edl with the ts file  this will skip from 1  minute to 2 minutes

```
60.00  120.00  3

emveepee:
@AlwinEsch  Updated to today's PR16485 changed to use GetEnum()  but the three c++ backend issues continue.   Do you need anything more from me?

Note I haven't updated my commit since I need @sub3  to commit 5.0.1 so I can rebase against what should be Matrix

AlwinEsch:
Have looked about one problem with recordings size. There you have a small mistake about the settings.

See my fixes here: https://github.com/AlwinEsch/pvr.nextpvr/commits/changes

But as another problem can't test size of recordings, as them given as Windows Path to my linux (Backend on Windows, Dev on Linux).
It brought: **D:\PVR-NEXTPVR\Hawaii Five-0\Hawaii Five-0_20200522_14021457.ts** and where it then add **smb:** on begin, but as "smb://D:/PVR-NEXTPVR/..." can not really work (maybe a set on settings to correct samba path about?).

Also another thing where confuse me here https://github.com/emveepee/pvr.nextpvr/blob/Matrix/src/BackendRequest.cpp#L108 as it use "127.0.0.1" (in original and C++ changes) (as note, the `#undef CreateDirectory` no more needed on addon).
**EDIT:** Them only to check about localhost or not?

AlwinEsch:
About the recordings -1 show, have added a patch to my Kodi PVR rework request, not sure if really correct but test has worked with and create also another request in Kodi about.

https://github.com/xbmc/xbmc/pull/16485/commits/a400818c79461c4af4301e568dd044b6703646a7

emveepee:
The reason recording size is an expert setting is to map it properly you would need your recordings  on a share and an advanced setting like this.  slash s preprocessed to backslash,

```
substitute>                                                                     
  <from>D:/PVR-NEXTPVR/</from>                                                  
  <to>smb://ip/sharename</to>                                            
</substitute>  
```
The hardcoded 127.0.0.1 is correct NextPVR interprets localhost differently for ipv6, but those few lines will get replaced with PingBackend. I will remove the undef CreateDirectory though

emveepee:
@AlwinEsch I didn't check the recent changes in PR16485 but i was able to get rid of the file size error by initializing to tag.SetSizeInBytes(0)

emveepee:
@AlwinEsch  I solved both remaining issue on comskip and set position.  In my conversion, I didn't set the proper start time for Recordings to 0.  
https://github.com/emveepee/pvr.nextpvr/blob/cpp-interface/src/buffers/RecordingBuffer.cpp#L16

Sorry to trouble you.  Now to rebase and finalize some testing.

emveepee:
@phunkyfish,  @AlwinEsch 
I have been able to open Shoutcast radio streams with Shoutcast metadata in Matrix using the standard OpenFile() method

https://github.com/emveepee/pvr.nextpvr/blob/cpp-interface/src/buffers/Buffer.cpp#L41

however with the C++ change there is no Shoutcast data and it appears that is because Shoutcast::Process() is never called.   If I change to use curl 
```
    m_inputHandle.CURLCreate(ss.str());
    m_inputHandle.CURLOpen();
```
I get the Shoutcast data.  In general is there any disadvantage to using curl instead of OpenFile() for all streams?

phunkyfish:
I didn’t think there was a difference.

So in your old code Using the old API you used OpenFile instead of CURLCreate/CURLOpen and it worked ok?

emveepee:
> 
> 
> I didn’t think there was a difference.
> 
> So in your old code Using the old API you used OpenFile instead of CURLCreate/CURLOpen and it worked ok?

Yes I never considered using the curl options before.  I haven't figured out where the Shoutcast::Process() thread is started but that seems to be the issue.  

I also tried with GetStreamProperties()  PVR_STREAM_PROPERTY_STREAMURL type and that works too.

phunkyfish:
Ok, let's see what @AlwinEsch has to say. Maybe there is a difference between the new and old APIs, or maybe a default setting.

AlwinEsch:
Confuse me as the old and new does exact the same:
Old Open: https://github.com/xbmc/xbmc/blob/master/xbmc/addons/interfaces/Addon/AddonCallbacksAddon.cpp#L323-L335
New Open: https://github.com/xbmc/xbmc/blob/master/xbmc/addons/interfaces/Filesystem.cpp#L740-L756

Which flags has you used?

Also in old is the CURLCreate/CURLOpen.

emveepee:
Good comment. I realize that I wasn't passing the option ADDON_READ_NO_CACHE to curl.  As soon as I did it failed.  I checked the old code and for some reason  I defined READ_NO_CACHE globally to 0 in this commit https://github.com/kodi-pvr/pvr.nextpvr/commit/ecc64584879cd1bceca7b50a9faa34bb1c1fe45b#diff-60f7e3708240135081fe3828dea97c93R80

I don't even know what the default file flag of 0 means looking at the table https://codedocs.xyz/AlwinEsch/kodi/group__cpp__kodi__vfs___defs___open_file_flags.html#gaa56e033cf2df273d11ee445768125d7c
it isn't defined.

This actually may turn out very positive, I do have users complaining about not being able to play recordings in progress even when the stream time are correct.  What flag is recommended for growing http files?

emveepee:
@phunkyfish I've spend most the of today looking at the timer and timer rules dialog and I  cannot get them to work in v5 (in C or C++)   On edits it is always defaulting to Any Channel and not showing the Recording Type.   The lists that I select edit from do have the correct type and channel

On new recordings the default shows Any Channel but the save sends one channel unless I actually select Any channel.   When the setting drop downs are opened the correct values are shown highlighted.   I did walk it through the debugger and the "filler" values in the settings menu are being selected they just aren't being shown on the dialog.

Also I can't create or edit pre or post padding values in EPG based rules.  They do show up on manual recordings.  Do you see any of this in v5?

phunkyfish:
I noticed something similar yesterday which I had not seen before with any channel vs a single channel. It was so long ago the last time I worked in this area I don’t quite remember it.

On padding I think you need to add the right flags to the timer type.

emveepee:
> 
> 
> I noticed something similar yesterday which I had not seen before with any channel vs a single channel. It was so long ago the last time I worked in this area I don’t quite remember it.
> 
I don't typically use Kodi for scheduling it was only after using Leia that I saw it working as expected that I thought about.

> On padding I think you need to add the right flags to the timer type.

The flags are the same for manual and guide based rules   PVR_TIMER_TYPE_SUPPORTS_START_END_MARGIN

https://github.com/kodi-pvr/pvr.nextpvr/blob/Matrix/src/Timers.cpp#L421

phunkyfish:
Is the padding the same on Leia? Maybe epg based recordings don’t use padding or it needs to be supported by the backend.

On both though it’s possible a bug was introduced.

emveepee:
> 
> Is the padding the same on Leia? Maybe epg based recordings don’t use padding or it needs to be supported by the backend.
> 
> On both though it’s possible a bug was introduced.

On Leia it works fine.   It could be a bug I introduced but I don't see where.

Ok for the padding the issue is here https://github.com/xbmc/xbmc/blob/master/xbmc/pvr/dialogs/GUIDialogPVRTimerSettings.cpp#L1234 and L1236 the backend  has added constraints that are incorrect for EPG based recordings.

Still digging why the channel default and type are not being showing the bestMatchingValue is being found.

phunkyfish:
> Ok for the padding the issue is here https://github.com/xbmc/xbmc/blob/master/xbmc/pvr/dialogs/GUIDialogPVRTimerSettings.cpp#L1234 and L1236 the backend has added constraints that are incorrect for EPG based recordings.

Why are they incorrect for epg based recordings?

> Still digging why the channel default and type are not being showing the bestMatchingValue is being found.

It’s almost like the logic behind the UI and the UI are not reflecting each other.

emveepee:
> 
> 
> > Ok for the padding the issue is here https://github.com/xbmc/xbmc/blob/master/xbmc/pvr/dialogs/GUIDialogPVRTimerSettings.cpp#L1234 and L1236 the backend has added constraints that are incorrect for EPG based recordings.
> 
> Why are they incorrect for epg based recordings?

NextPVR doesn't have the ability to edit the start or end time of EPG based recordings.  Those times are fixed to the times of the guide data.  Only the padding time is user editable.

emveepee:
> 
> It’s almost like the logic behind the UI and the UI are not reflecting each other.

The setting loads the list box correctly and selects the proper list items when the list dialog opens but it doesn't change the default text on the main dialog until opened.  I think it is related most to the type since I change the type to a new setting and back to the expected type the proper channel does change on the main dialog.

phunkyfish:
Testing this with vuplus I notice that if I create a new timer from the guide the default in the dialog is a once off guide based timer but if I go to change the timer type the default selection is an auto guide based timer.

phunkyfish:
There appears to be a mismatch somewhere and it's also there using the old C API.

emveepee:
> 
> 
> There appears to be a mismatch somewhere and it's also there using the old C API.

Yes I see it in both too.  I think we are seeing the text for the 0 value from the list for Type and Channel.  However if you do try and edit a timer and hit OK  you don't get the 0 value passed to the addon unless you actually select a new item.

phunkyfish:
So the timer UI is not being refreshed correctly by the data behind it.

I have no idea where to look for that.

emveepee:
> 
> 
> So the timer UI is not being refreshed correctly by the data behind it.
> 
> I have no idea where to look for that.

(This is the C interface) It seems related to the 0 on these lines https://github.com/xbmc/xbmc/blob/master/xbmc/pvr/dialogs/GUIDialogPVRTimerSettings.cpp#L229 and on L252 controlling the original text.

phunkyfish:
The timer is added here: https://github.com/xbmc/xbmc/blob/a288c97e191a45ff3c5c3bb54c6ec0bebff854cb/xbmc/pvr/guilib/PVRGUIActions.cpp#L408

And the timer settings are shown here: 
https://github.com/xbmc/xbmc/blob/a288c97e191a45ff3c5c3bb54c6ec0bebff854cb/xbmc/pvr/guilib/PVRGUIActions.cpp#L458

phunkyfish:
Might be best to create a kodi issue for this. I'm a little lost with timers.

emveepee:
@AlwinEsch  once PR 18027 https://github.com/xbmc/xbmc/pull/18027 is merged I can do a final test.   Other than that it should be ok to go subject to any review changes.

src/Channels.cpp
```diff
@@ -47,7 +45,7 @@ std::string Channels::GetChannelIcon(int channelID)
   std::string iconFilename = GetChannelIconFileName(channelID);
 
   // do we already have the icon file?
-  if (XBMC->FileExists(XBMC->TranslateSpecialProtocol(iconFilename.c_str()), false))
+  if (kodi::vfs::FileExists(kodi::vfs::TranslateSpecialProtocol(iconFilename), false))
```
AlwinEsch: Here you can try without "TranslateSpecialProtocol" normally should Kodi does it too.

emveepee: Yes there was a specific case, where I had to do this,  I will take them all out and see if I can find an example.   I also will change a remove() to FileDelete() to see if it works multi-platform

Also I will update to addon.cpp/h naming for consistency but to tell you the truth I liked client.cpp/h much better.

AlwinEsch: You can also leave as client.cpp/h, was only by me on others, as it is not direct the PVR client itself and client parts on other file.

emveepee: No I will change it I want to be the same it helps when comparing pvr addons knowing where to start.   That reminds me I also have to change addon.xml.in

Thinking of addon.xml it would be nice if a setting in addon.xml could disable or even hide Run on the addon screen for PVR addons.  I saw a post yesterday and I've also received a few posts that say the addon doesn't work when they click it.

emveepee: @AlwinEsch I don't need to use the translate so thanks for that recommendation but I still need to #undef CreateDirectory and DeleteFile or I get compile errors on Windows

'DeleteFileA': is not a member of 'kodi::vfs'	
'CreateDirectoryA': is not a member of 'kodi::vfs'

AlwinEsch: That confuse me, them I added here https://github.com/xbmc/xbmc/blob/master/xbmc/addons/kodi-addon-dev-kit/include/kodi/c-api/filesystem.h#L21-L28 and by my Windows tests it was OK.

Was then now need to add with "A" (DeleteFileA)? Think to take gun for Windows with this problems :smirk:, stupid that they have so much define conflicts.

phunkyfish: @emveepee  can you try including the filesystem include as the last header file and see does that make a difference?

emveepee: @phunkyfish changing header order makes a difference.  I narrowed it down further and when I swap the order of https://github.com/emveepee/pvr.nextpvr/blob/cpp-interface/src/BackendRequest.h#L12 to

#include "p8-platform/threads/mutex.h"
#include <kodi/Filesystem.h>

I don't need the #undef`

It must be related to this include of windows.h https://github.com/xbmc/platform/blob/master/src/windows/os-threads.h#L34

phunkyfish: You should not need to use any threading headers from P8s platform. All that should be able to come from the std library.

emveepee: I  guess I can change this legacy code to <mutex> and ignore the issue, but isn't Process() https://github.com/emveepee/pvr.nextpvr/blob/cpp-interface/src/pvrclient-nextpvr.cpp#L281 from P8 platform too?

Moving to `<mutex>` didn't help,  The issue is windows.h needs to get loaded before Kodi/Filesystem.h

phunkyfish: Process is just a function. I changed to std c++ threading in vuplus here:


https://github.com/kodi-pvr/pvr.vuplus/blob/Matrix/src/Enigma2.h

emveepee: Ok thanks I will leave that until after c++ I need to modify that loop anyway to allow async channel listing updates.

emveepee:
@AlwinEsch taking out the p8 platform "lost" the <unistd.h> header which explains the fail on linux, I was building in WIndows to test the windows.h issue.

AlwinEsch:
Has seen there was a bit mix about sleeps (std::chrono (1x), SLEEP macro (several times), without macro (1x)), 

Maybe use the C++11 std::chrono and thread on all places?
https://github.com/AlwinEsch/pvr.nextpvr/commit/348a529398343e7c05ca78fdc7ca0d956e207c34

emveepee:
> Has seen there was a bit mix about sleeps (std::chrono (1x), SLEEP macro (several times), without macro (1x)),
> 
> Maybe use the C++11 std::chrono and thread on all places?
> [AlwinEsch@348a529](https://github.com/AlwinEsch/pvr.nextpvr/commit/348a529398343e7c05ca78fdc7ca0d956e207c34)

Even that is not a complete solution because the p8 Sleep() is still used and I was going to look into this when I get rid of p8 thread after c++ is release.  How long before 16485 is released?  I might make the change in the base now.

AlwinEsch:
All addons are ready and mostly ready to come.
Have maybe a look here for yours https://jenkins.kodi.tv/job/BuildMultiWithAddons-PR/67/ and his OS logs

emveepee:
> 
> 
> All addons are ready and mostly ready to come.

Ok you seem so close I won't bother with the baseline code and just update everything to std::chrono

> Have maybe a look here for yours https://jenkins.kodi.tv/job/BuildMultiWithAddons-PR/67/ and his OS logs

I saw the fail but I didn't find the binary build logs, so I saw the error on a rebuild in linux which is why I submitted the update this morning.  Are the binary build logs available too?

emveepee:
@sub3  what version do you want me to bump this too?  Version 5 was the beginning of the major restructuring that formed the base for this so I can go to 5.1.0  or if you want it can go to 6.0.0  My preference is to stay on 5 since a jump from Leia v3 to Matrix v6  when we aren't even alpha with very few users seems odd.

phunkyfish:
My two cents are that this is a major change in the pvr api and so should be a major change in the addon version too.

That being said there are already a couple of major version in Matrix. @ksooo thoughts?

emveepee:
Given that @AlwinEsch  bumped all addons to PVR API 6.5.1 and right away for 7.0.0 I will join suit.  It's not worth the bandwidth either way.  Getting this approved, merged and tagged is what is important.

emveepee:
@AlwinEsch  let me know if there is something I missed.  The need for the SetEnumSettings commit is very odd, it didn't compile for Kodi on Windows but worked on other platforms.    The issue was/is it does compile and run fine for me on Windows and Linux so I would like to know what compile switch was causing the error (cannot cast float to <T>) on the Kodi side.

AlwinEsch (approved):

AlwinEsch:
Yes, that seen also on other addon's where use this helper functions. On Windows is a bit bad that by float it is against enum, as never called there. 

~~As alternative can further the `GetInt` used and interpret this to a enum.~~
**EDIT:** Forget it, my mind was wrong.

Make if possible a test with some other pvr and inputstream addons available.
Have brought the request in Kodi, where the `ATTRIBUTE_HIDDEN` is added on all parts.

Only to be sure that not part of addon A land in addon B.

emveepee:
The issue  isn't related to ATTRIBUTE_HIDDEN it turns out it is a compile difference between the cmake Windows Makefile and the  solution file used for a GUI compile.  I can generate the error with the Makefile.

Edit:  turns on the default cmake option for creating the VS solution file did not force the solution to the v141 toolset.  Adding  -T "v141" I get consistent error messages.

PR #130, opened 2020-05-28
5.0.2 READ_NO_CACHE

emveepee:
Addon was not passing the correct value for READ_NO_CACHE,  Modify  radio test and change behaviour of Recording and direct play URL's

Backport to Leia needed

### DIFF
--- a/src/buffers/ClientTimeshift.cpp
+++ b/src/buffers/ClientTimeshift.cpp
@@ -95,6 +95,24 @@ void ClientTimeShift::Close()
   m_lastClose = time(nullptr);
 }
 
+void ClientTimeShift::Resume()
+{
+  ClientTimeShift::GetStreamInfo();
+  if (m_stream_duration > m_settings.m_timeshiftBufferSeconds)
+  {
+    int64_t startSlipBuffer = m_stream_length - (m_settings.m_timeshiftBufferSeconds * m_stream_length / m_stream_duration);
+    XBMC->Log(LOG_DEBUG, "%s:%d: %lld %lld %lld", __FUNCTION__, __LINE__, startSlipBuffer, m_streamPosition, m_stream_length.load());
+    if (m_streamPosition < startSlipBuffer)
+    {
+      Seek(m_streamPosition, 0);
+    }
+  }
+  else
+  {
+    XBMC->Log(LOG_DEBUG, "%s:%d:", __FUNCTION__, __LINE__);
+  }
+}
+
 void ClientTimeShift::StreamStop()
 {
   std::string response;
--- a/src/buffers/ClientTimeshift.cpp
+++ b/src/buffers/ClientTimeshift.cpp
@@ -14,7 +14,7 @@
 
 using namespace timeshift;
 
-bool ClientTimeShift::Open(const std::string inputUrl, bool isRadio)
+bool ClientTimeShift::Open(const std::string inputUrl)
 {
   m_isPaused = false;
   m_stream_length = 0;
--- a/src/buffers/RecordingBuffer.cpp
+++ b/src/buffers/RecordingBuffer.cpp
@@ -16,7 +16,7 @@ PVR_ERROR RecordingBuffer::GetStreamTimes(PVR_STREAM_TIMES *stimes)
   stimes->startTime = 0;
   stimes->ptsStart = 0;
   stimes->ptsBegin = 0;
-  stimes->ptsEnd = ((int64_t ) Duration() ) * DVD_TIME_BASE;
+  stimes->ptsEnd = static_cast<int64_t>(Duration()) * DVD_TIME_BASE;
   return PVR_ERROR_NO_ERROR;
 }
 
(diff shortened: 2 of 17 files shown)


phunkyfish:
Looks good. Can you add the versioning info.

emveepee:
> 
> 
> Looks good. Can you add the versioning info.

Sure, I have commented out code that plays files  via the filesystem since we have on going issues with in progress especially http and I want to reenable this first, but I wanted @sub3 to test http.   

For recordings do you know what the value should be for ptsStart and ptsEnd for in progress recordings?  This READ_NO_CACHE issue was introduced in my attempts to try and make in progress more reliable and I haven't had much success based on user feedback.  Zero seems best for the start based, otherwise I get negative numbers or large numbers so the end seems to be the issue, but it could be both.

If I set the ptsEnd to the total time in microseconds there is no indication on the time line where the live point is so it is a bit confusing for the user they where it is in progress.    If I set end to the clock microseconds of playtime the time line is clickable and shows total time ok,  but with this is if a users hits stop near the "live" point the resume point isn't set because it is typically within <ignorepercentatend>#</ignorepercentatend>  of the end.

emveepee:
I will continue to use the v4 method for displaying in progress recordings but someday I would like to get answers on the correct method.  

@sub3  with v5 you must buffer a lot to memory without flush because in-progress files cannot use the file system.  Not a big deal http works fine as you tested and file system for completed recordings is not an issue.  I included the 3.2.20 resume change since I have to do a major rebase of the c++ change anyway.  Hopefully after @phunkyfish 's review this won't be held too for release too long.

sub3:
There no significant about of buffering being performed beyond what the filesystem is providing by default. ie, when it comes to the capture side, it's just writing the data to the file as it arrives. It doesn't explicitly flush data to disk - it just lets the filesystem take care of that when it's ready. 

Just let me know whey you're comfortable with it being merged and labelled.

src/buffers/ClientTimeshift.cpp
```diff
@@ -95,6 +95,24 @@ void ClientTimeShift::Close()
   m_lastClose = time(nullptr);
 }
 
+void ClientTimeShift::Resume()
+{
+  ClientTimeShift::GetStreamInfo();
+  if (m_stream_duration > m_settings.m_timeshiftBufferSeconds)
+  {
+    int64_t startSlipBuffer = m_stream_length - (m_settings.m_timeshiftBufferSeconds * m_stream_length/m_stream_duration);
```
phunkyfish: Spaces before and after `/`

phunkyfish (changes requested):

src/buffers/ClientTimeshift.cpp
emveepee: Fixed.

emveepee:
@phunkyfish  thanks for your feedback.  One note on MenuHooks, since I could find no way to clear existing MenuHooks there is a minor hole in the logic.  If a user has Kodi sleeping and changes the version of  the backend, on resume the MenuHooks are based on the original connection version.

phunkyfish:
Do you have connection established logic in this addon?

emveepee:
> Do you have connection established logic in this addon?

Yes it disconnects on sleep and reconnects on wake to avoid session security timeouts.  This PR addressed MenuHooks being readded as they were being created during the connect logic but now they need to be independent of the reconnect logic.

phunkyfish:
> > Do you have connection established logic in this addon?
> 
> Yes it disconnects on sleep and reconnects on wake to avoid session security timeouts. This PR addressed MenuHooks being readded as they were being created during the connect logic but now they need to be independent of the reconnect logic.

Ah, ok. I will admit I'm not familiar with them ;)

emveepee:
>
> Ah, ok. I will admit I'm not familiar with them ;)

Just thought I'd mention it since you have a keen eye reviewing changes so you might spot something in other addons.  Thanks as always for the help

src/buffers/RecordingBuffer.cpp
```diff
@@ -16,7 +16,7 @@ PVR_ERROR RecordingBuffer::GetStreamTimes(PVR_STREAM_TIMES *stimes)
   stimes->startTime = 0;
   stimes->ptsStart = 0;
   stimes->ptsBegin = 0;
-  stimes->ptsEnd = ((int64_t ) Duration() ) * DVD_TIME_BASE;
+  stimes->ptsEnd = ((int64_t)Duration()) * DVD_TIME_BASE;
```
phunkyfish: C++ style cast?

src/pvrclient-nextpvr.h
```diff
@@ -89,6 +89,7 @@ class cPVRClientNextPVR : P8PLATFORM::CThread
   bool CanSeekStream(void);
   bool IsTimeshifting(void);
   bool IsRealTimeStream(void);
+  bool IsRadio() {return m_nowPlaying == Radio;};
```
phunkyfish: Space before return and before closing bracket for clarity.

Like this: `{ return m_nowPlaying == Radio; };`

phunkyfish (changes requested):

src/buffers/RecordingBuffer.cpp
emveepee: Changed

src/pvrclient-nextpvr.h
emveepee: Thanks for the tip.  Changed.   I also got to practice amending an earlier commit using the method you gave me on another PR, well worth knowing.

phunkyfish (approved):
Nice, looks good 😉

emveepee:
@sub3 this can be tagged and released and I can then rebase PR 128 and we should be good with the c++ conversion

PR #132, opened 2020-06-09
5.0.3  Timer changes

emveepee:
Start and end times not being set correctly on any time type 1 and 2.    Set the original time on timeslot recording.

### DIFF
--- a/pvr.nextpvr/addon.xml.in
+++ b/pvr.nextpvr/addon.xml.in
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <addon
   id="pvr.nextpvr"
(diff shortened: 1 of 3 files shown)


emveepee:
@sub3 when you have a chance after approval from @phunkyfish could you tag and merge this.  In Matrix this is critical because of a crash,  Somewhere Matrix is doing a local time to UTC conversion  with a  local time of 0  which is failing in UTC+ timezones.  This logic improvement avoids the root cause of the crash.  I will also port this to Leia as it is a more accurate representation but Leia is not crashing.

Note that after the C++ port I will make the rules names in Kodi look more like the v5 web.

phunkyfish (approved):

PR #133, opened 2020-06-17
6.0.1 ConnectionStateChange

emveepee:
After the c++ conversion, the ConnectionStateChange() message parameter must be a string,  and an empty string is needed to use the default string.  Menuhook was always being attached even when the client didn't connect


phunkyfish (approved):
LGTM

emveepee:
@sub3  @phunkyfish  approved this so I think it should be tagged and merged when you have time.  It is a crashing bug if the server is not available because of the expansion to nullptr.c_str()

PR #134, opened 2020-06-21
6.0.2 In-progress recording improvement

emveepee:
Use in-progress change logic tested in Leia. Fix default port definition.

### DIFF
--- a/src/buffers/RecordingBuffer.cpp
+++ b/src/buffers/RecordingBuffer.cpp
@@ -25,12 +25,17 @@ int RecordingBuffer::Duration(void)
   if (m_recordingTime)
   {
     std::unique_lock<std::mutex> lock(m_mutex);
-    time_t endTime =  time(nullptr);
-    int diff = (int) (endTime - m_recordingTime -10);
+    int diff = static_cast<int>(time(nullptr) - m_recordingTime) - 15;
     if (diff > 0)
+    {
       m_isLive = true;
+      diff += 15;
+    }
     else
+    {
       m_isLive = false;
+      diff = 0;
+    }
     return diff;
   }
   else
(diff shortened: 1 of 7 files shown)


src/buffers/RecordingBuffer.cpp
```diff
@@ -79,11 +84,15 @@ int RecordingBuffer::Read(byte *buffer, size_t length)
   if (dataRead == 0 && m_isLive)
   {
     kodi::Log(ADDON_LOG_DEBUG, "%s:%d: %lld %lld", __FUNCTION__, __LINE__, m_inputHandle.GetLength() , m_inputHandle.GetPosition());
-    int64_t position = m_inputHandle.GetPosition();
-    Buffer::Close();
-    Buffer::Open(m_recordingURL, ADDON_READ_NO_CACHE);
-    Seek(position, 0);
-    dataRead = (int) m_inputHandle.Read(buffer, length);
+    const int64_t position = m_inputHandle.GetPosition();
+    const time_t startTime = time(nullptr);
```
phunkyfish, line 88: Just. a comment. The const is not really required here as both variables are integral types. You can still use it to indicate this variable should be changed but generally it gets used for complex types. Just a comment though and entirely up to you.

src/buffers/RecordingBuffer.cpp
```diff
@@ -79,11 +84,15 @@ int RecordingBuffer::Read(byte *buffer, size_t length)
...
+    do {
+      Buffer::Close();
+      std::this_thread::sleep_for(std::chrono::milliseconds(2000));
+      Buffer::Open(m_recordingURL);
+      Seek(position, 0);
+      dataRead = static_cast<int>(m_inputHandle.Read(buffer, length));
+    } while (dataRead == 0 && time(nullptr) - startTime < 5);
```
phunkyfish, line 95: Do you want to calculate time(nullptr) on each loop or would once before the loop be sufficient?

phunkyfish (commented):
Only comments ;)

src/buffers/RecordingBuffer.cpp
emveepee: On each loop, I don't want to wait more than 5 seconds, and in case at real EOF the read doesn't return -1 I want to exit.

phunkyfish: Ah, sorry, makes sense. NVM

src/buffers/RecordingBuffer.cpp
emveepee: Ok I didn't know that I thought Kodi want const for checks on function calls.  I will keep this in for now

phunkyfish (approved):

phunkyfish (approved):

PR #135, opened 2020-06-28
Timeshift transcoding

emveepee:
Experimental attempt to use ffmpeg to timeshift live TV

### DIFF
--- a/src/pvrclient-nextpvr.cpp
+++ b/src/pvrclient-nextpvr.cpp
@@ -334,7 +334,7 @@ PVR_ERROR cPVRClientNextPVR::OnSystemWake()
   int count = 0;
   for (; count < 5; count++)
   {
-    if (Connect())
+    if (Connect() == ADDON_STATUS_OK)
     {
       g_pvrclient->ConnectionStateChange("connected", PVR_CONNECTION_STATE_CONNECTED, "");
       break;
--- a/src/pvrclient-nextpvr.cpp
+++ b/src/pvrclient-nextpvr.cpp
@@ -98,6 +98,15 @@ cPVRClientNextPVR::cPVRClientNextPVR(const CNextPVRAddon& base, KODI_HANDLE inst
 
 cPVRClientNextPVR::~cPVRClientNextPVR()
 {
+  if (m_nowPlaying != NotPlaying)
+  {
+    // this is likley only needed for transcoding but include all cases
+    if (m_nowPlaying == Recording)
+      CloseRecordedStream();
+    else
+      CloseLiveStream();
+  }
+
   StopThread();
 
   kodi::Log(ADDON_LOG_DEBUG, "->~cPVRClientNextPVR()");
@@ -200,6 +209,28 @@ void cPVRClientNextPVR::ConfigurePostConnectionOptions()
   {
     delete m_timeshiftBuffer;
     m_supportsLiveTimeshift = true;
+
+    if (m_settings.m_liveStreamingMethod == eStreamingMethod::Transcoded && m_settings.m_transcodedTimeshift)
+    {
+      std::string version;
+      bool enabled;
+      const std::string addonName = "inputstream.ffmpegdirect";
+
+      if (kodi::IsAddonAvailable(addonName, version, enabled))
+      {
+        if (!enabled)
+        {
+          kodi::Log(ADDON_LOG_INFO, "%s installed but not enabled at startup", addonName.c_str());
+          kodi::QueueFormattedNotification(QueueMsg::QUEUE_ERROR, kodi::GetLocalizedString(30191).c_str(), addonName.c_str());
+        }
+      }
+      else // Not installed
+      {
+        kodi::Log(ADDON_LOG_INFO, "%s not installed", addonName.c_str());
+        kodi::QueueFormattedNotification(QueueMsg::QUEUE_ERROR, kodi::GetLocalizedString(30192).c_str(), addonName.c_str());
+      }
+    }
+
     if (m_settings.m_liveStreamingMethod == eStreamingMethod::Transcoded)
     {
       m_supportsLiveTimeshift = false;
@@ -430,6 +461,12 @@ PVR_ERROR cPVRClientNextPVR::GetChannelStreamProperties(const kodi::addon::PVRCh
     {
       kodi::Log(ADDON_LOG_ERROR, "Transcoding Error");
       return PVR_ERROR_FAILED;
+    }
+    if (m_settings.m_transcodedTimeshift)
+    {
+      properties.emplace_back(PVR_STREAM_PROPERTY_INPUTSTREAM, "inputstream.ffmpegdirect");
+      properties.emplace_back("inputstream.ffmpegdirect.stream_mode", "timeshift");
+      properties.emplace_back("inputstream.ffmpegdirect.manifest_type", "hls");
     }
     properties.emplace_back(PVR_STREAM_PROPERTY_STREAMURL, line);
     properties.emplace_back(PVR_STREAM_PROPERTY_ISREALTIMESTREAM, "true");
--- a/pvr.nextpvr/addon.xml.in
+++ b/pvr.nextpvr/addon.xml.in
@@ -1,10 +1,12 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <addon
   id="pvr.nextpvr"
-  version="6.0.2"
+  version="6.0.3"
   name="NextPVR PVR Client"
   provider-name="Graeme Blackley">
-  <requires>@ADDON_DEPENDS@</requires>
+  <requires>@ADDON_DEPENDS@
+    <import addon="inputstream.ffmpegdirect" minversion="1.15.0" optional="true"/>
+  </requires>
   <extension
     point="kodi.pvrclient"
     library_@PLATFORM@="@LIBRARY_FILENAME@"/>
--- a/pvr.nextpvr/changelog.txt
+++ b/pvr.nextpvr/changelog.txt
@@ -1,6 +1,7 @@
 v6.0.3
 - Optionally use timeshifted transcoded streaming via inputstream.ffmpegdirect
 - Fix for incorect OnSystemWake loop exit after c++ conversion change
+- The prebuffer slider was not matched with the correct selection
 
 v6.0.2
 - Fix default port
--- a/pvr.nextpvr/changelog.txt
+++ b/pvr.nextpvr/changelog.txt
@@ -1,5 +1,6 @@
 v6.0.3
 - Optionally use timeshifted transcoded streaming via inputstream.ffmpegdirect
+- Fix for incorect OnSystemWake loop exit after c++ conversion change
 
 v6.0.2
 - Fix default port
--- a/pvr.nextpvr/changelog.txt
+++ b/pvr.nextpvr/changelog.txt
@@ -1,3 +1,6 @@
+v6.0.3
+- Optionally use timeshifted transcoded streaming via inputstream.ffmpegdirect
+
 v6.0.2
 - Fix default port
 - In-progress recordings change
--- a/pvr.nextpvr/resources/language/resource.language.en_gb/strings.po
+++ b/pvr.nextpvr/resources/language/resource.language.en_gb/strings.po
@@ -32,6 +32,10 @@ msgctxt "#30003"
 msgid "Enable timeshift with live TV"
 msgstr ""
 
+msgctxt "#30603"
+msgid "Open settings dialog for inputstream.ffmpegdirect for modification of timeshift settings"
+msgstr ""
+
 msgctxt "#30004"
 msgid "Download TV guide artwork"
 msgstr ""
@@ -276,4 +280,12 @@ msgstr ""
 
 msgctxt "#30190"
 msgid "Timeshifting disabled"
+msgstr ""
+
+msgctxt "#30191"
+msgid "%s is not enabled"
+msgstr ""
+
+msgctxt "#30192"
+msgid "%s is not installed"
 msgstr ""
--- a/pvr.nextpvr/resources/settings.xml
+++ b/pvr.nextpvr/resources/settings.xml
@@ -169,6 +169,17 @@
             </dependency>
           </dependencies>
         </setting>
+        <setting help="30603" id="ffmpegdirect" label="30003" type="boolean"  parent="livestreamingmethod5">
+          <level>1</level>
+          <default>false</default>
+          <control type="toggle"/>
+          <dependencies>
+            <dependency type="visible">
+              <condition operator="is" setting="livestreamingmethod5">3</condition>
+              <condition operator="is" setting="legacy">false</condition>
+            </dependency>
+          </dependencies>
+        </setting>
       </group>
       <group id="10">
         <setting help="" id="chunklivetv" label="30167" type="integer">
--- a/pvr.nextpvr/resources/settings.xml
+++ b/pvr.nextpvr/resources/settings.xml
@@ -101,7 +101,7 @@
           </constraints>
           <dependencies>
             <dependency type="visible">
-              <condition operator="is" setting="livestreamingmethod">2</condition>
+              <condition operator="is" setting="livestreamingmethod">1</condition>
               <condition operator="is" setting="legacy">true</condition>
             </dependency>
           </dependencies>
@@ -164,7 +164,7 @@
           </control>
           <dependencies>
             <dependency type="visible">
-              <condition operator="is" setting="livestreamingmethod5">2</condition>
+              <condition operator="is" setting="livestreamingmethod5">0</condition>
               <condition operator="is" setting="legacy">false</condition>
             </dependency>
           </dependencies>
--- a/src/Settings.cpp
+++ b/src/Settings.cpp
@@ -119,7 +119,7 @@ ADDON_STATUS Settings::ReadBackendSettings()
       }
 
       if (XMLUtils::GetInt(settingsDoc.RootElement(), "SlipSeconds", m_timeshiftBufferSeconds))
-        kodi::Log(ADDON_LOG_INFO, "time shift buffer in seconds == %d\n", m_timeshiftBufferSeconds);
+        kodi::Log(ADDON_LOG_INFO, "time shift buffer in seconds: %d", m_timeshiftBufferSeconds);
 
       std::string serverMac;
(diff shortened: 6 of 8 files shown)


emveepee:
@phunkyfish  could you please help me and let me know what additional steps I might need to do to enable timeshifting on an m3u8 file.  Running  with this PR I don't see timeshift controls on live tv.
```
2020-06-28 10:59:43.377 T:16748    INFO <general>: CAddonMgr::ADDON::CAddonMgr::FindAddons: inputstream.ffmpegdirect v1.14.5 installed
```

src/pvrclient-nextpvr.cpp
```diff
@@ -431,8 +431,10 @@ PVR_ERROR cPVRClientNextPVR::GetChannelStreamProperties(const kodi::addon::PVRCh
       kodi::Log(ADDON_LOG_ERROR, "Transcoding Error");
       return PVR_ERROR_FAILED;
     }
+    properties.emplace_back(PVR_STREAM_PROPERTY_INPUTSTREAM, PVR_STREAM_PROPERTY_VALUE_INPUTSTREAMFFMPEG);
```
phunkyfish: This needs to be: 
`properties.emplace_back(PVR_STREAM_PROPERTY_INPUTSTREAM, "inputstream.ffmpegdirect");`

You have set it to use kodi's internal ffmpeg inputstream and not the ffmpegdirect addon.

phunkyfish: You may also need to set these two properties:

`
properties.emplace_back("inputstream.ffmpegdirect.is_realtime_stream", "true");
properties.emplace_back("inputstream.ffmpegdirect.manifest_type", "hls");
`

Although it may work without them.

emveepee: Thanks a lot I didn't know there where two but now it makes sense.

Unfortunately that crashes almost immediately after playback in void CDVDDemuxUtils::FreeDemuxPacket(DemuxPacket* pPacket)  The Addon does create a 0 byte seg/idx file.

Here is the debug log https://paste.kodi.tv/ecisuyafij.kodi

phunkyfish: Can you provide a backtrace with a debugger?

phunkyfish: Can you provide a backtrace with a debugger?

emveepee: Does this help?
>
```
 	kodi.exe!av_packet_free_side_data()	C
>	kodi.exe!CDVDDemuxUtils::FreeDemuxPacket(DemuxPacket * pPacket) Line 32	C++
	kodi.exe!CInputStreamAddon::cb_free_demux_packet(void * kodiInstance, DemuxPacket * packet) Line 663	C++
```

phunkyfish: Anyway you can get me access to a test stream?

emveepee: Would the actual set of files help  https://1drv.ms/u/s!Ar-w3NVk_f-MlUxPnO67Uv7BVd4e?e=tOWCp0   A transcoded file is created on the fly by the nextpvr backend.

The logs for the thread are here https://paste.kodi.tv/mojusoyazi.kodi

emveepee: Actually cool it is working now with the nightly,  my self debug compile of kodi might be the issue, I will check the dependencies.

It works great but here are two issues I am having

1) - perhaps I am missing a setting I see lots of spam with

2020-06-29 10:11:47.264 T:54400   DEBUG <CAddonSettings[inputstream.ffmpegdirect]>: loading setting definitions

2) - the arrow  button doesn't take me to the beginning of the buffer, it takes me to random place.  I can skip on the timeline and with the mouse and cursor.   https://youtu.be/6CZ0zYB9WfU

phunkyfish: 1. That's not the addon. It's the debug code in the settings manager which probably needs to be fixed.

2. Yes, that's correct. The left arrow will try to take you to the start of the programme which is not possible. It works as expected once the next programme starts. I tried to add a correction there for the seek but no matter what I tried I couldn't get it to work.

emveepee: > 
>     2. Yes, that's correct. The left arrow will try to take you to the start of the programme which is not possible. It works as expected once the next programme starts. I tried to add a correction there for the seek but no matter what I tried I couldn't get it to work.

Do you set the SetStartTime() to clock time on the first play?

I think this is a really nice option for users of NextPVR that need transcoding remote and have the system space.  It is the next  best thing to skipping on directly via the m3u8 file.  Thanks for this feature and the help getting it going.

phunkyfish: Yes, it’s a different structure in an inputstream but similar to `GetStreamTimes` in PVR.

https://github.com/xbmc/inputstream.ffmpegdirect/blob/2d485c3b3a3db982be81466bc34df07508cb978d/src/stream/TimeshiftStream.cpp#L125

emveepee: The windows crashing problem no longer is an issue after compiling a debug version of the inputsream.ffmpegdirect so this  is resolved.

emveepee:
The PR has been updated with a settings option to enable ffmpegdirect for timeshifted transcoding.   Currently the addon install doesn't force any inputstream to be installed but inputstream.ffmpeg comes preinstalled on some platforms.

phunkyfish:
You can optionally install the inputstream if you would like from your addon xml:

https://github.com/kodi-pvr/pvr.iptvsimple/blob/8ff23d326f0ca6932d4b44ee5c845c0a732ab48f/pvr.iptvsimple/addon.xml.in#L7-L11

emveepee:
> 
> 
> You can optionally install the inputstream if you would like from your addon xml:
> 
> https://github.com/kodi-pvr/pvr.iptvsimple/blob/8ff23d326f0ca6932d4b44ee5c845c0a732ab48f/pvr.iptvsimple/addon.xml.in#L7-L11

Thanks I was wondering how to this, but it is an edge use case.  The transcoding feature is typically only for users who either want reduced bandwidth for remote access or who use devices that can't direct play mpeg2video, h264 and potential h265 and have backend powerful enough to do transcoding.   Do you know if linux will need to get these via the ppa?  Also is that the right minimum version for timeshifting and binary addons?

After looking at the iptvsimple code I did see I get test for the addon at runtime, that should do what I want.  In testing if inputstream.ffmpegdirect is not installed it will fallback to the original transcoded input stream

@sub3 what do you think ?

phunkyfish:
You can just set the minimum to the latest if you want. You may not even need to set a minimum. Maybe try with a min and see does it install?

Ubuntu would be from the PPA.

emveepee:
@phunkyfish is  ffmpegdirect timeshifting supposed to allow skipping to the beginning of the buffer live tranditional live TV?  I can't seem to go earlier then the current show.

phunkyfish:
I think you can’t go back to the a partial previous programme but you can go back to any full programme. But I’m not sure this isolated to ffmpegdirect.

emveepee:
> 
> 
> I think you can’t go back to the a partial previous programme but you can go back to any full programme. But I’m not sure this isolated to ffmpegdirect.

Here is a  video https://www.youtube.com/watch?v=4Ikwi0pd3gk showing it working fine in regular NextPVR timeshifting.  The left arrow goes to beginning of current show, prev start partial show, and right goes to next show, then end of show (live) as expected.

phunkyfish:
Then I stand corrected !

emveepee:
> 
> 
> Then I stand corrected !

No worries, I just want to make sure it wasn't something on my end.  When I release this to users I will remind them this is experimental and to hit pause.  It could be frustrating if the phone rings or they take a beer break, and come back after the show ends to find they can't see the end of the show.

phunkyfish:
It should be fixable. Probably just an edge case I didn’t check. I will take a look on Monday.

phunkyfish:
I remember now that when I saw this it was a pain to figure out as I always needed to have a programme boundary to test it. So unfortunately I never did.

EDIT: I just remembered what I found. A seek call is never passed to the addon when you seek using the prev button. So it’s an issue in core that needs to be fixed. Just don’t really have a clue how or where to start!

emveepee:
> EDIT: I just remembered what I found. A seek call is never passed to the addon when you seek using the prev button. So it’s an issue in core that needs to be fixed. Just don’t really have a clue how or where to start!

TimeshiftStream::DemuxSeekTime() is getting negative number with the backwards flag.  If I set that to zero rather than return false the left arrow works better and Seek() is called.  It still doesn't allow skipping to the old show but I think you need to handle negative seeks relative to something.

phunkyfish:
Very good. A fresh set of eyes always helps.

So that allows us to get back to the start of the current programme easily. Let me see if I can get seeking to the initial programme working.

phunkyfish:
Prev button should be fixed in this testbuild: https://jenkins.kodi.tv/blue/organizations/jenkins/xbmc%2Finputstream.ffmpegdirect/detail/PR-59/3/artifacts

PR is here: https://github.com/xbmc/inputstream.ffmpegdirect/pull/59

emveepee:
> 
> 
> Prev button should be fixed in this testbuild: https://jenkins.kodi.tv/blue/organizations/jenkins/xbmc%2Finputstream.ffmpegdirect/detail/PR-59/3/artifacts
> 
> PR is here: [xbmc/inputstream.ffmpegdirect#59](https://github.com/xbmc/inputstream.ffmpegdirect/pull/59)

Nice.  I can now seek back several and forth through programs as expected.   I was holding off on announcing this for NextPVR experimenting and no I have no concerns.

I do still have the same crash with the release ffmpegdirect build against debug Kodi on Windows but a debug build works fine.   

Now if https://github.com/xbmc/inputstream.ffmpegdirect/blob/Matrix/src/stream/FFmpegLog.cpp#L77 could changed to use a variable (and maybe be not quite as dynamic) to get rid of that millisecond spam this would be perfect.

phunkyfish:
> I do still have the same crash with the release ffmpegdirect build against debug Kodi on Windows but a debug build works fine.

I have no idea what causes this. I don’t use windows at all but a bit stuck as to how I would get it to work.

> Now if https://github.com/xbmc/inputstream.ffmpegdirect/blob/Matrix/src/stream/FFmpegLog.cpp#L77 could changed to use a variable (and maybe be not quite as dynamic) to get rid of that millisecond spam this would be perfect.

Believe me I tried. Maybe it’s just my `C` intermixed with `C++` which is poor but I couldn’t get this working with a variable.

It might be easier to try and fix the issue in kodi by removing that log statement wherever it comes from.

phunkyfish:
Actually, maybe a variable would be better. If you have any ideas on how to solve that one I'd be greatful!

emveepee:
> 
> 
> Actually, maybe a variable would be better. If you have any ideas on how to solve that one I'd be greatful!

Maybe someting like this.  I noticed you can't configure the logging while it is running anyway so needlessly loading the settings.xml so often makes less sense.

```
diff --git a/src/stream/FFmpegLog.cpp b/src/stream/FFmpegLog.cpp
index aa020c9..e2caec1 100644
--- a/src/stream/FFmpegLog.cpp
+++ b/src/stream/FFmpegLog.cpp
@@ -29,6 +29,21 @@ void FFmpegLog::SetLogLevel(int level)
   CFFmpegDirectLogTls = log;
 }
 
+void FFmpegLog::SetEnabled(bool enabled)
+{
+  FFmpegLog* log = CFFmpegDirectLogTls;
+  if (log)
+    log->enabled = enabled;
+}
+bool FFmpegLog::GetEnabled()
+{
+  FFmpegLog* log = CFFmpegDirectLogTls;
+  if (!log)
+    return false;
+  return log->enabled;
+}
+
+
 int FFmpegLog::GetLogLevel()
 {
   FFmpegLog* log = CFFmpegDirectLogTls;
@@ -74,7 +89,7 @@ void ff_avutil_log(void* ptr, int level, const char* format, va_list va)
   if (FFmpegLog::GetLogLevel() > 0)
     maxLevel = AV_LOG_INFO;
 
-  if (level > maxLevel && !kodi::GetSettingBoolean("allowFFmpegLogging"))
+  if (level > maxLevel && !FFmpegLog::GetEnabled())
     return;
 
   LogLevel type;
diff --git a/src/stream/FFmpegLog.h b/src/stream/FFmpegLog.h
index 98add67..7619afa 100644
--- a/src/stream/FFmpegLog.h
+++ b/src/stream/FFmpegLog.h
@@ -27,9 +27,12 @@ class FFmpegLog
 {
 public:
   static void SetLogLevel(int level);
+  static void SetEnabled(bool enabled);
+  static bool GetEnabled();
   static int GetLogLevel();
   static void ClearLogLevel();
   int level;
+  bool enabled;
 };
 
 } //namespace ffmpegdirect
diff --git a/src/stream/FFmpegStream.cpp b/src/stream/FFmpegStream.cpp
index 8987bcc..2359d65 100644
--- a/src/stream/FFmpegStream.cpp
+++ b/src/stream/FFmpegStream.cpp
@@ -144,7 +144,7 @@ bool FFmpegStream::Open(const std::string& streamUrl, const std::string& mimeTyp
   m_mimeType = mimeType;
   m_isRealTimeStream = isRealTimeStream;
   m_programProperty = programProperty;
-
+  FFmpegLog::SetEnabled(kodi::GetSettingBoolean("allowFFmpegLogging"));
   if (m_openMode == OpenMode::CURL)
     m_curlInput->Open(m_streamUrl, m_mimeType, ADDON_READ_TRUNCATED |
                                                ADDON_READ_BITRATE |
```

phunkyfish:
Great, I will test this tomorrow. Thanks!

phunkyfish:
I tried your patch but `GetEnabled()` always seem to evaluate to false so I never see any ffmpeg logging.

emveepee:
> 
> 
> I tried your patch but `GetEnabled()` always seem to evaluate to false so I never see any ffmpeg logging.

Did you change the location where it runs?  There does need to be a log started to use this.  

image

phunkyfish:
Ya, `FFmpegLog::SetLogLevel(AV_LOG_INFO);` is called in the constructor so it should exist. But clearly it's not working as I expect. I think this might be the issue I came across last time I tried to use a var for this.

Or did you mean something else?

emveepee:
> 
> 
> Ya, `FFmpegLog::SetLogLevel(AV_LOG_INFO);` is called in the constructor so it should exist. But clearly it's not working as I expect. I think this might be the issue I came across last time I tried to use a var for this.
> 
> Or did you mean something else?

I basically wanted to know if you run the patch on PR59 or on modified code.  It clearly is working for me which is why I should the debugger breakpoint

phunkyfish:
I ran the patch on the PR. It applied perfectly. It's possible behaviour is different on mac osx vs windows. But I have no idea why.

emveepee:
> 
> 
> I ran the patch on the PR. It applied perfectly. It's possible behaviour is different on mac osx vs windows. But I have no idea why.

I spent the afternoon on my "mac" and eventually I was able to download compile and install ffmpegdirect  Here are my logs http://paste.kodi.tv/pihejidiro   If you search for "Dll Initializing - Inputstream FFmpeg Direct" the first set is with the logging enabled the second is with it disabled.  There sure seems to be a difference to me.

phunkyfish:
Yes. That is working exactly as expected. Hmmm. Let me do a full rebuild on my side because there must be something wring with my setup.

src/pvrclient-nextpvr.cpp
emveepee: All issues resolved with https://github.com/xbmc/inputstream.ffmpegdirect/pull/59

emveepee:
@phunkyfish  this is ready for review and when 1.15.0 is released this can go.  I take back all comments about not being able to play the source m3u8 file.  The transcoding option in NextPVR is really designed for use when direct play bandwidth is too high, so a local cache is perfect.

Edit:  I take that back.  I can't seem to install the zip file with the changed addon.xml.in  that you suggested.  The install still fails to install because it seems to be required.

```
2020-07-15 14:51:28.780 T:14736   DEBUG <general>: CAddonInstaller: installing from zip 'D:\Downloads\pvr.nextpvr-6.0.3.zip'
2020-07-15 14:51:28.794 T:11020   DEBUG <general>: CAddonInstallJob[pvr.nextpvr]: requires inputstream.ffmpegdirect version 0.0.0 which is not available
2020-07-15 14:51:28.794 T:11020   ERROR <general>: CAddonInstallJob[pvr.nextpvr]: The dependency on inputstream.ffmpegdirect version 0.0.0 could not be satisfied.
```

Could you review my addon.xml.in as well not sure why it is looking for 0.0.0.0 and why it fails to install anyway

this is what shows in the created addon.xml

```
  <requires>
<import addon="kodi.binary.global.main" minversion="1.2.0" version="1.2.4"/>
<import addon="kodi.binary.global.general" minversion="1.0.4" version="1.0.5"/>
<import addon="kodi.binary.global.gui" minversion="5.14.0" version="5.14.1"/>
<import addon="kodi.binary.global.filesystem" minversion="1.1.0" version="1.1.4"/>
<import addon="kodi.binary.global.network" minversion="1.0.0" version="1.0.4"/>
<import addon="kodi.binary.instance.pvr" minversion="7.0.0" version="7.0.1"/>
    <import addon="inputstream.ffmpegdirect" minversion="1.15.0" optional="true"/>
  </requires>

```

phunkyfish:
That's only because 1.15.0 was not released yet!

Wait about an hour and it should work then.

phunkyfish (approved):
Nice, looks good.

emveepee:
> 
> 
> That's only because 1.15.0 was not released yet!
> 
> Wait about an hour and it should work then.

Ok but if it is optional shouldn't it let the install continue with it  missing?     Edit ignore that, I checked the Kodi code and the optional flag is only if the addon is not installed.  I uninstalled 1.1.45 and I could install the update to pvr.nextpvr

phunkyfish:
That’s only if the version is valid.

phunkyfish:
For you travis error you should use the C++11 way instead of using LONG_MAX.

Include: `<limits>`
`std::numeric_limits<long>::max()`

emveepee:
> 
> 
> For you travis error you should use the C++11 way instead of using LONG_MAX.
> 
> Include: `<limits>`
> `std::numeric_limits<long>::max()`

Great even better I can use std::numeric_limits<time_t>::max(); and remove my comment about 2038  Thanks a lot.

PR #137, opened 2020-07-16
[Matrix] increase version to 7.0.0 (API related) and cleanups

AlwinEsch:
Request related to API update about GUI here xbmc/xbmc#18408

Need merge here after Kodi.


AlwinEsch:
The GUI parts is swapped to here and with the warning fixes from before.
Think the GUI comes in next days and not need a wait like inputstream/pvr part and his GSOC project.

PR #138, opened 2020-07-23
6.1.0 Enhancement

emveepee:
Add optional season on recordings,  pass through genres.  Don't lock out addon on network error.

### DIFF
--- a/pvr.nextpvr/resources/language/resource.language.en_gb/strings.po
+++ b/pvr.nextpvr/resources/language/resource.language.en_gb/strings.po
@@ -297,3 +297,11 @@ msgstr ""
 msgctxt "#30693"
 msgid "Create separate folders for multiple seasons of recordings"
 msgstr ""
+
+msgctxt "#30194"
+msgid "Use backend genre text"
+msgstr ""
+
+msgctxt "#30694"
+msgid "Kodi default is DVB genre text.  Default required for Kodi genre search and language translation"
+msgstr ""
--- a/pvr.nextpvr/resources/language/resource.language.en_gb/strings.po
+++ b/pvr.nextpvr/resources/language/resource.language.en_gb/strings.po
@@ -183,7 +183,7 @@ msgid "Kodi skin styled recordings"
 msgstr ""
 
 msgctxt "#30569"
-msgid "Separate season and episode for skin - Forced them in subtitle when off (default)"
+msgid "Separate title, season and episode for skin - season and episode numbers will not be shown in lists"
 msgstr ""
 
 msgctxt "#30170"
@@ -231,11 +231,11 @@ msgid "Remote access"
 msgstr ""
 
 msgctxt "#30180"
-msgid "Single file display"
+msgid "Single recording display"
 msgstr ""
 
 msgctxt "#30680"
-msgid "Movies and single episode TV shows not shown in folders"
+msgid "Movies and single episode TV shows display at top level"
 msgstr ""
 
 msgctxt "#30181"
@@ -288,4 +288,12 @@ msgstr ""
 
 msgctxt "#30192"
 msgid "%s is not installed"
+msgstr ""
+
+msgctxt "#30193"
+msgid "Show season folders"
+msgstr ""
+
+msgctxt "#30693"
+msgid "Create separate folders for multiple seasons of recordings"
 msgstr ""
--- a/src/Recordings.cpp
+++ b/src/Recordings.cpp
@@ -185,7 +185,7 @@ bool Recordings::UpdatePvrRecording(TiXmlElement* pRecordingNode, kodi::addon::P
   }
 
   // v4 users won't see recently concluded recordings on the EPG
-  int endEpgTime = g_pvrclient->XmlGetInt(pRecordingNode, "epg_end_time_ticks");
+  int endEpgTime = XMLUtils::GetIntValue(pRecordingNode, "epg_end_time_ticks");
   if (endEpgTime > time(nullptr) - 24 * 3600)
   {
     tag.SetEPGEventId(endEpgTime);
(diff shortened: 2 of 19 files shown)


emveepee:
It is not clear why all the Windows failed with tinyxml even when the Jenkins and Travis builds are fine.  Also it seems the Travis information which Ia cleaned up a bit for this build doesn't seem to be on the build but I found it here.  https://travis-ci.org/github/kodi-pvr/pvr.nextpvr/builds/713485382  I expect come more will be fixed up with PR 141 and the change to hiddenn

I am hoping that this can be released soon.  Some users are already testing the earlier artifacts and I'd like to start moving to tinyxml-2 now before adding some new backend calls for v5.  @sub3  @phunkyfish  could you please review.

phunkyfish:
There is something up with the azure pipelines currently. Not sure what is. Tagging @AlwinEsch for reference.

emveepee:
@AlwinEsch I did another PR today with the same azure pipeline issues.

AlwinEsch:
Have created a issue here https://developercommunity.visualstudio.com/content/problem/1132859/failure-about-cstrawberrycbinpatchexe-since-202007.html, hopefully they fix it soon.

emveepee:
@sub3 @phunkyfish  any chance for a review?  I'd really like to see this merged and tagged and then I can fix up the tinyxml-2 PR and move forward.

pvr.nextpvr/resources/language/resource.language.en_gb/strings.po
```diff
@@ -289,3 +289,19 @@ msgstr ""
...
+msgid "Create separate folders for multiple seasons of recordings"
+msgstr ""
+
+msgctxt "#30194"
+msgid "Use backend genre text"
+msgstr ""
+
+msgctxt "#30694"
+msgid "Kodi default is DVB genre text.  Default required for Kodi genre search and language translation"
+msgstr ""
```
AlwinEsch: One place I see and only a missing new line

src/Settings.cpp
```diff
@@ -142,6 +148,24 @@ ADDON_STATUS Settings::ReadBackendSettings()
...
+void Settings::SetConnection(bool status)
+{
+  if (status == true)
+  {
+      kodi::vfs::CFile outputFile;
+      if (outputFile.OpenFileForWrite(connectionFlag))
+      {
+        outputFile.Close();
+      }
+      m_connectionConfirmed = true;
```
AlwinEsch, line 157: About here and only as idea, it can also make a bit smaller (and without the extra 2 spaces ;) )
```cpp
  if (status == true)
  {
    kodi::vfs::CFile outputFile;
    m_connectionConfirmed = outputFile.OpenFileForWrite(connectionFlag);
  }
```
The Close not direct needed as done by his destructor.

The Close can also be called, but more for cases as CFile is in class values or after read/write comes bigger work where takes time.

src/Recordings.cpp
```diff
@@ -296,58 +335,66 @@ bool Recordings::UpdatePvrRecording(TiXmlElement* pRecordingNode, kodi::addon::P
...
-        const char* clear = node->Attribute("clear");
-        if (strStringValue.empty() || (clear && StringUtils::CompareNoCase(clear, "true") == 0))
-          strStringValue = strTemp;
-        else
-          strStringValue += strSeparator + strTemp;
+
+        std::ssub_match base_sub_match = base_match[1];
+        tag.SetSeriesNumber(std::stoi(base_sub_match.str()));
+        base_sub_match = base_match[2];
+        tag.SetEpisodeNumber(std::stoi(base_sub_match.str()));
```
AlwinEsch, line 355: About `std::stoi` I'm not sure, Maybe the `atoi`?
Heard from @phunkyfish that this is not safe. I'm right?

AlwinEsch (commented):
Can not direct test, but what I see looks it good.

AlwinEsch:
About the Azure problem I found the reason, see https://github.com/xbmc/xbmc/pull/18247
They have there now cmake 3.18.0 and this breaks the current Kodi.

AlwinEsch:
One part, seen there the commit names `"Connection state change #1"` and `"Connection state change #2"` them link against requests.

With `#` makes wrongly links to them and send also maybe mails to places where not need.

emveepee:
> 
> 
> About the Azure problem I found the reason, see [xbmc/xbmc#18247](https://github.com/xbmc/xbmc/pull/18247)
> They have there now cmake 3.18.0 and this breaks the current Kodi.

I upgraded to 3.18.1 and do see the same issue so I reverted to  3.17.3 which is clean.   The change probably requires I rebuild the base from scratch so I will wait for this PR to merge to make sure it is clean.    

Confirmed fixed now with cmake 3.18.1 as a new install

pvr.nextpvr/resources/language/resource.language.en_gb/strings.po
emveepee: Fixed

emveepee:
> 
> 
> On part, seen there the commit names `"Connection state change #1"` and `"Connection state change #2"` them link against requests.
> 
> With `#` makes wrongly links to them and send also maybe also mails to places where not need.

Ok removed the # I didn't know this.  I have two more coming after the tinyxml-2 merge too

src/Settings.cpp
emveepee: Ok changed but I set the state to true even if the write fails because at this point it has connected to the backend.

phunkyfish:
I can’t review till next week as I’m away.

emveepee:
> 
> 
> I can’t review till next week as I’m away.

Thanks I will ask @sub3 to review and release this thanks  users are using the artifacts anyway.  

If you  have time could you address the comment on using std::stoi  in general, it looks like the change that was noted was automated since I don't see the values in c to c++ just to use stoi but I will never use it if it is wrong.

src/Recordings.cpp
phunkyfish: Stoi won’t return 0 if it can’t read a numeric value. So you may want atoi instead.

emveepee: > 
> 
> Stoi won’t return 0 if it can’t read a numeric value. So you may want atoi instead.

Ah I see now I would need try/catch  in some cases.   Most of them are safe, including this one because they follow regex \d but this is important to know.  Thanks both of you. 

Edit none of the other std::stoi calls in the addon should be unsafe unless the backend sends non numeric values in error, but I will clean this up in the tinyxml-2 conversion since tinyxml-2 is using c style strings anyway and atoi makes sense.

PR #139, opened 2020-07-23
Get rid of external kodiplatform dependency

basilgello:
1. import XMLUtils.h from pvr.vuplus and change
   "namespace enigma2" to "namespace NextPVR"

2. Modify the includes with script:

HEADER_PATH=$(find $PWD/ -name XMLUtils.h)
git grep -l "XMLUtils.h" | while read FILE
do
  RELATIVE_PATH=$(realpath --relative-to=$(dirname $(readlink -f "$FILE")) "$HEADER_PATH")
  sed -i 's|["<]kodi/util/XMLUtils.h[>"]|"'$RELATIVE_PATH'"|' "$FILE"
done

3. Open the affected files manually and re-roder includes:

   nano $(git grep -l "XMLUtils\.h")

   Also add "using namespace NextPVR::utilities" and replace
   "XMLUtils::" with "xml::"

4. Remove libkodiplatform-dev form debian/control file

5. Remove kodiplatform fromCMakeLists.txt and add XMLUtils.h
   to local header list.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -2,7 +2,7 @@ Source: kodi-pvr-nextpvr
 Priority: extra
 Maintainer: Nobody <nobody@kodi.tv>
 Build-Depends: debhelper (>= 9.0.0), cmake, libtinyxml-dev,
-               libkodiplatform-dev (>= 16.0.0), kodi-addon-dev
+               kodi-addon-dev
 Standards-Version: 4.1.2
 Section: libs
 Homepage: http://www.nextpvr.com
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,12 +4,10 @@ project(pvr.nextpvr)
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${PROJECT_SOURCE_DIR})
 
 find_package(Kodi REQUIRED)
-find_package(kodiplatform REQUIRED)
 find_package(p8-platform REQUIRED)
 find_package(TinyXML REQUIRED)
 
 include_directories(${p8-platform_INCLUDE_DIRS}
-                    ${kodiplatform_INCLUDE_DIRS}
                     ${TINYXML_INCLUDE_DIR}
                     ${KODI_INCLUDE_DIR}/..) # Hack way with "/..", need bigger Kodi cmake rework to match right include ways
 
@@ -57,10 +55,10 @@ set(NEXTPVR_HEADERS src/addon.h
                     src/buffers/RecordingBuffer.h
                     src/buffers/CircularBuffer.h
                     src/buffers/RollingFile.h
-                    src/buffers/Seeker.h)
+                    src/buffers/Seeker.h
+                    src/utilities/XMLUtils.h)
 
 SET(DEPLIBS ${p8-platform_LIBRARIES}
-            ${kodiplatform_LIBRARIES}
             ${TINYXML_LIBRARIES})
 if(WIN32)
   list(APPEND DEPLIBS ws2_32)
--- a/pvr.nextpvr/addon.xml.in
+++ b/pvr.nextpvr/addon.xml.in
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <addon
   id="pvr.nextpvr"
-  version="6.0.3"
+  version="6.0.4"
   name="NextPVR PVR Client"
   provider-name="Graeme Blackley">
   <requires>@ADDON_DEPENDS@
--- a/pvr.nextpvr/changelog.txt
+++ b/pvr.nextpvr/changelog.txt
@@ -1,3 +1,6 @@
+v6.0.4
+- Get rid of external kodiplatform dependency
+
 v6.0.3
 - Optionally use timeshifted transcoded streaming via inputstream.ffmpegdirect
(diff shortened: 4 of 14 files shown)


emveepee:
Isn't it better to use the platform library rather then one under addon control?  If they vuplus one is better shouldn't it update the platform library.

phunkyfish:
We have been slowly removing the library from the addons. Vuplus was first.

Early on the intention was to build out kodi platform to perform other functions but that never happened so better to remove it.

basilgello:
> Isn't it better to use the platform library rather then one under addon control? If they vuplus one is better shouldn't it update the platform library.

My goal is to remove the kodiplatform package from Debian after I asked @phunkyfish about tagged release of it and he replied that's likely gonna be removed from Kodi along with p8-platform. Removing p8 is a bit tough for now but kodiplatorm is easy.

emveepee:
Ok thanks I looked for a master PR.  I really was hoping for the opposite move though where we could get CDate and other useful Kodi functions from kodiplatform and also use the full XMLUtils rather then each addon having to do that.

phunkyfish:
If it was possible to have header only versions of the functionalities you want sure. But as a library we would prefer to stop using it. It is overhead only used by a few addons so should be removed.

AlwinEsch:
Some functions can added to Kodi's addon headers if needed.

About XMLUtils and his TinyXML, one reason why I tought to kick out kodi-platform was that TinyXML since 5 years no more maintained and makes then not direct sense to continue his use with kodi-platform.

There are already on few addons, other maintained alternatives used, tinyxml2, pugixml, libxml2

emveepee:
I would move to TinyXML-2 if it was an option but overall doesn't make moving XMLUtils into the addon makes harder not easier if you want all addons do it.  That to me is what platform is all about.

phunkyfish:
Tinyxml2 is completely different to tinyxml. 

But the platform shouldn’t be dictating an xml lib anyway. That should be up to the implementation.

emveepee:
I don't see the harm in making it easier for addon devs to share common dependencies to do standard things like xml processing.  If the common library has problem it is updated and addon dev may not even need to make a change.  If the library is moved to the addon there is more risk and more work so I don't see a real advantage.  It is only less of a problem now because tinyxml is static.  

I don't think of this as Kodi "dictating" that I use it either.    I have been considering moving to tinyxml2 anyway because of the large number of channels users have so extra efficiency would be good.

phunkyfish:
And this is what many other addons have done already. What was a common well used library is now only used by a few addons. Times have changed and there are better options available.

I went with pugixml on iptvsimple cause it’s c++11.

basilgello:
I think the

> Date and other useful Kodi functions from kodiplatform

should be exposed as headers in kodi addon SDK as Alwin suggested.

As for XML processing library, there are two choices: first, leave it up to developer (current situation) or select the best-maintained and performing library and wrap the SDK around it making that SDK a part of kodi-addons-dev (kodi addon SDK).

The second solution will require defacto design of the whole wrapper SDK from scratch and making Kodi using it and also efforts to port all legacy stuff to the new SDK but will eliminate dozens of same-purpose libraries in the process memory.

Anyway, kodiplatform and p8-platform seem to provide overlapping functionality on kodi addon SDK and the unification effort is welcome here!

AlwinEsch:
Also the StringUtils part of p8-platform (where taken from Kodi itself) are in next time no more really needed.
With C++20 becomes the most used part added and can be used for e.g. string format.

Only what I find a bit stupid currently that Raspberry still takes gcc 4 and we can not use for now :unamused:, hope there comes a update to use newer compiler versions.

AlwinEsch:
Also I'm not a fan to use in library another lib and where using target must work with both API's about (thats why I not so liked the kodi-platform). 
In this cases would then better for me to get a own API about and where the used lib on addon speak only with them.
e.g. like Qt

basilgello:
This is what I mean by providing true platform: generic CML parsers, exposed specific tasks like M3U/EPG/CUE parsing etc.

AlwinEsch (approved):

phunkyfish (approved):
👍

emveepee:
The PR is much nicer now without xml:: thank you for that.  @sub3 is not available could someone else merge and tag this if it is ready?

AlwinEsch:
Has brought it in and released, thanks a lot for this changes!

debian/control
```diff
@@ -2,7 +2,7 @@ Source: kodi-pvr-nextpvr
 Priority: extra
 Maintainer: Nobody <nobody@kodi.tv>
 Build-Depends: debhelper (>= 9.0.0), cmake, libtinyxml-dev,
-               libkodiplatform-dev (>= 16.0.0), kodi-addon-dev
+               kodi-addon-dev
```
AlwinEsch, line 5: Shit, there then need to add `libp8-platform-dev`, Sorry that I'm not seen before :sweat_smile:.

Can you add here on your other request the `libp8-platform-dev`?

AlwinEsch: I think to look how debian build can be included inside travis or AppVeyor.

Trying now request about.

basilgello: @AlwinEsch I built this patch without p8 using debian-development VM (git-pbuilder etc)

p8 is needed for StringUtils.h ?

EDIT: Ups... sorry, I also mixed two addons! Will the PR ASAP.

PR #140, opened 2020-07-24
debian: add libp8-platform-dev as build-dep

basilgello:
* kodiplatform-dev has libp8-platform-dev as
   build-dependency so it was fetched by gbp.

@AlwinEsch


phunkyfish:
This will need a version bump as well.

phunkyfish (approved):

phunkyfish:
Also released.

PR #142, opened 2020-07-31
Conversion to tinyxml-2

emveepee:
Test build for tinyxml-2  Mirror all backend calls to see that returned xml parses correctly.  Replace one function to see how it works.    The intent of the PR will be to replace tinyxml .


emveepee:
Ok my test build is good so this should be a straight forward upgrade.  @AlwinEsch  @phunkyfish  In another thread there were issues identified that tinyxml is not being maintained but I see that Kodi only has an older version of tinyxml-2  http://mirrors.kodi.tv/build-deps/sources/     How do I get a newer version included in build-deps?

Can there anything more confusing then the version numbers for tinyxml and tinyxml2? V1 is tinyxml-2.6.2_2.tar.gz and v2 is tinyxml2-6.2.0.tar.gz

AlwinEsch:
You can use the Link from here https://github.com/leethomason/tinyxml2/releases/tag/8.0.0 / https://github.com/leethomason/tinyxml2/archive/8.0.0.tar.gz

Can also upload it to our mirror, but by github URL's not direct needed as them fast enough.

AlwinEsch:
About versioning, yes thats there really stupid.

emveepee:
> 
> 
> You can use the Link from here https://github.com/leethomason/tinyxml2/releases/tag/8.0.0 / https://github.com/leethomason/tinyxml2/archive/8.0.0.tar.gz
> 
> Can also upload it to our mirror, but by github URL's not direct needed as them fast enough.

Ok thanks I assume I just put the URL in depends with the correct sha256 It seems to be working.

I thought the version number originally was bad.  The github file that gets downloaded is just 8.0.0.tar.gz

emveepee:
Ok, I converted everything to tinyxml2 and removed tinyxml from the build and everything seems in order.  @AlwinEsch thanks for the help, the  azure pipeline error continues on tinyxml2.

@phunkyfish  could you please review.  After you review PR 138 I will ask @sub3 to merge it and then resolve conflicts with this PR, update the changelog and merge both as 6.1.0

emveepee:
> 
> You can use the Link from here https://github.com/leethomason/tinyxml2/releases/tag/8.0.0 / https://github.com/leethomason/tinyxml2/archive/8.0.0.tar.gz

@AlwinEsch I saw that tinxyxml-dev was in debian/control and took it out but I don't know how to test that is it correct now, could you review?  I also removed kodi-platform since I assume that was only there for tinxyxml.

phunkyfish:
pvr.vbox uses tinyxml2.

emveepee:
> 
> 
> pvr.vbox uses tinyxml2.

Yes but it just includes old source files not using find etc.   pvr.dvblink is closer and it uses find but it also adds libtinyxml2-dev to the debian/control file and I don't know if that is needed or not.

phunkyfish:
I would assume it is required. There would need to be some reference to tiny xml 2 in that file.

emveepee:
> 
> 
> I would assume it is required. There would need to be some reference to tiny xml 2 in that file.

My concern is @AlwinEsch  suggested 8.0.0 and that version of libtinyxml2-dev isn't even exist in Ubuntu 20.04.  It seems to be compile without it.  Alternatively I can revert to the older kodi version that probably is more supported.

I notice there is also a line in the debian/control file that maybe I need to add 

Recommends: inputstream-ffmpegdirect

phunkyfish:
Unless you know if the API changes in 8.0.0 don't effect what you use it would be safer to the use a version that is at least compatible with 20.04.

I'm sure that version 7 will be new enough.

emveepee:
> 
> I'm sure that version 7 will be new enough.

Seems there was a critical bug in 7.0.1 with homebrew so I updated this to 7.0.1  

This is one reason I feel it is a good idea to keep common libraries in kodi-platform.   3 pvr'ssknow use different versions of tinyxml2 and Kodi uses another.

phunkyfish:
I meant to use the latest stable version of v7.

Having this as a common library would be worse. Then for every bump each addon would need to be retested. Very hard to do when those addons may not have maintainers.

emveepee:
> 
> I meant to use the latest stable version of v7.
> 
Bumped to 7.1.0

phunkyfish:
@emveepee why remove the Doxygen comments in XMLUtils.h?

Leaving them there (and possible updating them) makes it far more likely your work on this will get reused by other addons ;)

emveepee:
> 
> 
> @emveepee why remove the Doxygen comments in XMLUtils.h?
> 
> Leaving them there (and possible updating them) makes it far more likely your work on this will get reused by other addons ;)

As it is I don't think it is a good base.  From the first commit pvr.nextpvr was loose mixing node and element names which has always confused me and I had to change some calls.  I will look at cleaning that up first and if I can get rid of the ambiguity and convert nodes to nodes and elements to elements then I can add the comments back.

Edit it should be closer now, probably better to use  new style doxygen

src/Channels.cpp
```diff
@@ -29,8 +29,8 @@ int Channels::GetNumChannels()
       tinyxml2::XMLDocument doc;
       if (doc.Parse(response.c_str()) == tinyxml2::XML_SUCCESS)
       {
-        tinyxml2::XMLElement* channelsNode = doc.RootElement()->FirstChildElement("channels");
-        tinyxml2::XMLElement* pChannelNode;
+        tinyxml2::XMLNode* channelsNode = doc.RootElement()->FirstChildElement("channels");
```
phunkyfish, line 32: After looking at the tinyxml2 docs I'm not sure I understand switching to the use of XMLNode. A Node is simply a base class for many other types. What is the benefit of this vs using XMLElement?

FYI, everything that was read into a type of XMLElement is in fact an XMLElement so I'm not sure I see the benefit here.

emveepee: > 
> 
> FYI, everything that was read into a type of XMLElement is in fact an XMLElement so I'm not sure I see the benefit here.

It was simply because all the variable names say  "Node"  I felt it might be confusing to others who aren't used as I am that might want to use this XMLUtils.h.  Also the original Doxygen referenced nodes not elements so I wanted to be consistent with the documentation I readded.

phunkyfish: Ok, regardless of what it is named in the code everywhere it is used should be an XMLElement and not an XMLNode. So I would change the variable names and not the type of you wish.

emveepee: > 
> 
> Ok, regardless of what it is named in the code everywhere it is used should be an XMLElement and not an XMLNode. So I would change the variable names and not the type of you wish.

Why isn't it better to keep it the same as core, https://github.com/xbmc/xbmc/blob/master/xbmc/utils/XMLUtils.h and the old library in kodi-platform?

phunkyfish: Sorry, I thought your version was the same as core. I didn’t realise you had changed it.

I understand now.

emveepee:
Final review for release requested.

src/BackendRequest.h
```diff
@@ -40,15 +40,17 @@ namespace NextPVR
...
-    void clearSid() { std::fill(std::begin(m_sid), std::end(m_sid), 0); m_sidUpdate = 0; };
+    tinyxml2::XMLError  GetLastUpdate(const char* resource, time_t& last_update);
     bool PingBackend();
     bool OneTimeSetup();
-    const char* getSID() { return m_sid; };
+    const char* getSID() { return m_sid.c_str(); };
     std::vector<std::vector<std::string>> Discovery();
-    bool isSidActive() { return time(nullptr) < m_sidUpdate + 900; };
+
+    void setSID(std::string newsid) { m_sid = newsid; };
```
phunkyfish, line 51: `SetID`

src/BackendRequest.h
```diff
@@ -40,15 +40,17 @@ namespace NextPVR
...
+    void clearSid() { m_sid.clear(); m_sidUpdate = 0; };
```
phunkyfish, line 52: `ClearSid`

src/BackendRequest.h
```diff
@@ -40,15 +40,17 @@ namespace NextPVR
...
+    bool isSidActive() { return !m_sid.empty() && time(nullptr) < m_sidUpdate + 900; };
```
phunkyfish, line 53: `IsSidActive`

src/Recordings.cpp
```diff
@@ -124,7 +125,20 @@ PVR_ERROR Recordings::GetRecordings(bool deleted, kodi::addon::PVRRecordingsResu
   {
     returnValue = PVR_ERROR_SERVER_ERROR;
   }
-  g_pvrclient->m_lastRecordingUpdateTime = time(0);
+  g_pvrclient->m_lastRecordingUpdateTime = time(nullptr);
+  return returnValue;
+}
+PVR_ERROR Recordings::GetRecordingsLastPlayedPosition()
```
phunkyfish, line 131: New line above

src/pvrclient-nextpvr.h
```diff
@@ -108,8 +108,9 @@ class ATTRIBUTE_HIDDEN cPVRClientNextPVR : public kodi::addon::CInstancePVRClien
   MenuHook& m_menuhook = MenuHook::GetInstance();
   Recordings& m_recordings = Recordings::GetInstance();
   Timers& m_timers = Timers::GetInstance();
-  int64_t m_lastRecordingUpdateTime;
-  time_t m_nextServerCheck = 0;
+  time_t m_lastRecordingUpdateTime;
+  time_t m_nextServerCheck{ 0 };
```
phunkyfish, line 106: Do you want this field to be set to zero or now?

src/BackendRequest.h
emveepee: I am setting the URL parameter &sid= so I believe SetSID is better for my understanding but I see what you mean SetSID()

src/BackendRequest.h
emveepee: Ok same things but I don't like CamelCase for short abbreviations so ClearSID()

src/BackendRequest.h
emveepee: Ok I will adjust them all to be consistent.

src/pvrclient-nextpvr.h
emveepee: Technically I would want this to be zero but  it is only checked  when it is set 60 seconds after  a connection failure.  It is basically for Kodi style.  Does this set it to now?

phunkyfish: That’s what I’m not sure about. As it’s based  on a basic type I would just always use `= 0;`

src/BackendRequest.h
emveepee: Next release SID functions updated

src/pvrclient-nextpvr.h
emveepee: Checked with the debugger and it is 0 which makes sense.  I changed all time initializations to = 0 for the next release

src/Recordings.cpp
emveepee: Changed for next release

PR #143, opened 2020-08-24
Update pvrclient-nextpvr.cpp

dev77-07:
Returns the total and used drive space for GetDriveSpace

### DIFF
--- a/src/pvrclient-nextpvr.cpp
+++ b/src/pvrclient-nextpvr.cpp
@@ -416,8 +416,35 @@ PVR_ERROR cPVRClientNextPVR::GetConnectionString(std::string& connection)
 
 PVR_ERROR cPVRClientNextPVR::GetDriveSpace(uint64_t& total, uint64_t& used)
 {
-  total = 0;
-  used = 0;
+  std::string response;
+  int64_t spaceTotal = 0;
+  int64_t spaceUsed = 0;
+  if (m_request.DoRequest("/service?method=system.space", response) == HTTP_OK)
+  {
+    TiXmlDocument doc;
+    if (doc.Parse(response.c_str()) != nullptr)
+    {
+      TiXmlElement* spaceNode = doc.RootElement()->FirstChildElement("directory");
+      kodi::Log(ADDON_LOG_INFO, "spaceNode %d", spaceNode);
+      if (spaceNode != nullptr)
+      {
+        while (spaceNode)
+        {
+          int64_t tmpTotal;
+          int64_t tmpFree;
+          XMLUtils::GetLong(spaceNode, "total", tmpTotal);
+          XMLUtils::GetLong(spaceNode, "free", tmpFree);
+          int64_t tmpUsed = tmpTotal - tmpFree;
+          spaceTotal += tmpTotal;
+          spaceUsed += tmpUsed;
+
+		  spaceNode = spaceNode->NextSiblingElement("directory");
+        }
+      }
+    }
+  }
+  total = spaceTotal / 1024;
+  used = spaceUsed / 1024;
   return PVR_ERROR_NO_ERROR;
 }
 


emveepee:
Thanks for the contribution.   I had several reason for not implementing this based on my testing.

- Kodi continuously spams the server with this call when you are browsing or just sitting in the recording library.  Please send the debug pastbin logs of your implementation sitting on the recordings screen and I will show you what I mean if you don't see it.   I would suggest returning variables that are set only when recording updates are triggered.  Not perfect but better then the spam

- the return can be exceptionally slow calculating space on a network drive or waking a green drive.

- you can't arbitrarily sum the space.  If you have several recording  folders LiveTV, Movies and TVShows there is no indication they are on different drives.  Ideally the API should be based on drives not folders.  The fear is over reporting free space.

- if you have one drive that is almost full and one drive that has lots of free space, there is not indication that the recording scheduled on the nearly full drive could fail.  I have this real world scenario with a small SSD as one of my drives.

I'm not sure if @sub3  addressed any of those points in the API because the web server calls them too.  Perhaps he could comment.  

Also did you test against a v4 backend or should this be a v5 feature only.  Not sure the if the API changed but I suggest limiting it to v5 anyway to be safe.

That all being said I am not against it as an option if  you can reduce the spam.  Would it be hard for you to add a config option to turn it off and on?  I not I would be happy to do that part

phunkyfish:
Kodi will only repeatedly call the addon for this. The addon can choose to contact the backend as often/seldom as it likes.

For instance. The addon could choose to only call the backend once a minute.

emveepee:
> 
> 
> Kodi will only repeatedly call the addon for this. The addon can choose to contact the backend as often/seldom as it likes.

To clarify for me Kodi means core+addon.  I am asking @yeahme49  to come up with a strategy that doesn't request this update every 1.5 seconds

Ideally it would be preferable if the backend API returned something more consistent with the Kodi requirement, however for situations where the user only as one storage folder configured it is valid which is why I believe having an optional configuration would be of benefit to a few users.  Perhaps changes on the backend could do some caching to avoid going to the file system as often as it does.  There is nothing invalid about having a recording folder that is temporarily offline,  but the timeout for a lookup is over 20 seconds.

@sub3  the bottom right part of the screen is how Kodi displays this information as you may not have seen it before.  This information would technically be better on a timer page since it is recording capacity, which may not even be related to  disk space used by the recordings.

image

emveepee:
I thought I'd show my setup to give you an idea of the concern over summing directories as drives.  In the web app it doesn't look bad common sense helps.   

image

My prime reason for monitoring is to see my Live TV and Default space which does go to near zero if I don't pay attention.  The other two drives (technically three but Networks and Archive are one drive on a NAS with two share points) and I still need to check this manually.

Certainly we can make a simple key so that if total:free is the same we skip it for the total and I think this is a good idea.

The second problem is the bigger I keep some files on folder on a share that is not always onine and .  The 25 seconds is just too long on a blocking call from core and there is no good time to initialize it. Similar times are expected if USB drivers are in  the mix.  

```
2020-08-25 10:56:35.018 [DEBUG][144]    Got request [::1]: /service (system.space)
2020-08-25 10:56:35.018 [DEBUG][144]    method=system.space
2020-08-25 10:56:35.018 [DEBUG][144]    parameters:
2020-08-25 10:56:35.018 [DEBUG][144]       method: system.space
2020-08-25 10:56:35.018 [DEBUG][144]       client_ip: ::1
2020-08-25 10:56:35.018 [DEBUG][144]       user_agent: curl/7.55.1
2020-08-25 10:56:35.018 [DEBUG][144]       host_callback: localhost:8866
2020-08-25 10:56:35.018 [DEBUG][144]       sid: default
2020-08-25 10:57:01.876 [DEBUG][144]    Unexpected error checking available disk space on: \\172.16.3.2\Videos
2020-08-25 10:57:01.876 [ERROR][144]    Unexpected error finding available space: System.IndexOutOfRangeException: Index was outside the bounds of the array.
   at NShared.Extend.SystemRequestHandler.GetSpace(Dictionary`2 parameters)
2020-08-25 10:57:01.876 [DEBUG][144]    <?xml version="1.0" encoding="utf-8" ?>
<rsp stat="failed">
</rsp>
```

The fail is cached but @sub3 would need to clarify when the 

Here is the API returns this for an offline drive

```
   <directory name="Network">
      <total>100</total>
      <free>0</free>

**Comment by sub3** [comment]
@yeahme49 - thanks for the PR. I think this was a reasonable attempt at gaining this functionality. For most users, they're probably only using a single recording folder, and this will give them the intended result. 

As described by emveepee, there are complications, like the user having multiple recording directories (some possibly pointing to the same drive, some to network drives which might not be online etc).  Ultimately, it's difficult to map these on to the more simplistic "GetDriveSpace" call that Kodi uses.

Given how frequently Kodi can call this, and how the data will not make sense in all setups, and because some volumes like network drives can be slow to return this information, I think we'd at least need a setting to be added, so this function could be enabled/disabled before it's merged in to the main branch. 

In the NextPVR version of NextPVR, I'll add some logic to cache this info, so it's not queried too frequently.

**Comment by sub3** [comment]
Actually thinking about it further, as well as making it optional, it probably needs to have some client side logic to throttle the frequency of the backend call. ie, cache the result, and only make new requests every few minutes, otherwise we're going to get the NextPVR logs full of this system.space call.

**Comment by emveepee** [comment]
This is going to fail with 6.2+ because tinyxml is gone.  You can use the new examples based on DeMethodRequest() to simplify the call.

```
  tinyxml2::XMLDocument doc;
  if (request.DoMethodRequest("/service?method=system.space", doc) == tinyxml2::XML_SUCCESS)
  {
      // use doc  directly

Also now that things works better in disconnected state this should only be called when  

m_connectionState == PVR_CONNECTION_STATE_CONNECTED

Otherwise there will be a 30 second timeout if core calls it.

emveepee:
@yeahme49 any progress or thoughts?  @sub3 if you really want the feature added let me know and I can submit an alternate PR.

emveepee:
@sub3 unless you have any objection could you close this PR.  @yeahme49 if you'd like could you review https://github.com/kodi-pvr/pvr.nextpvr/pull/153  where I believe I have addressed my concerns on this PR.  Feel free to test as well.

sub3:
This is implemented in #153, so closing this one.

PR #144, opened 2020-08-28
Updates and clean up

emveepee:
Fix client timeshift plus refactor code

### DIFF
--- a/pvr.nextpvr/addon.xml.in
+++ b/pvr.nextpvr/addon.xml.in
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <addon
   id="pvr.nextpvr"
-  version="6.2.0"
+  version="6.2.1"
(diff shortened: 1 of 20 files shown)


emveepee:
@phunkyfish  @sub3 There was a regression in the implementation fixing the IPTV issue (time reported by backennd clock time not a potentially must longer time from reading the m3u8) causing quick tuners to fail so I would like this out fairly soon.   I am not sure I buffer enough in the current implementation and will want to get user feedback.  I might have to revert to the initial implementation where I only trust clock time.

phunkyfish (approved):
LGTM

emveepee:
@phunkyfish  @sub3  I decide to rework and simplify the timeshift buffer it is now primarily based on time.

- if no data is delivered in 20 seconds abort  
- if any data is available allow backend to buffer live tv the user defined time 0 or 1 seconds is typically enough

PR #146, opened 2020-09-08
Protect recording and live buffers

emveepee:
PVR Recordings cannot use ADDON_OPEN_CACHED even when finalized.  Kodi core can issue calls against streams that are not open Kodi issue https://github.com/xbmc/xbmc/issues/18368

### DIFF
--- a/pvr.nextpvr/resources/settings.xml
+++ b/pvr.nextpvr/resources/settings.xml
@@ -29,6 +29,14 @@
             <heading>30002</heading>
           </control>
         </setting>
+        <setting help="" id="hostprotocol" label="" type="string" option="hidden">
+          <level>4</level>
+          <default>http</default>
+          <control format="string" type="edit">
+            <heading></heading>
+            <hidden>true</hidden>
+          </control>
+        </setting>
       </group>
       <group id="8">
         <setting help="" id="wolenable" label="30163" type="boolean">
--- a/src/pvrclient-nextpvr.cpp
+++ b/src/pvrclient-nextpvr.cpp
@@ -137,7 +137,6 @@ ADDON_STATUS cPVRClientNextPVR::Connect(bool sendWOL)
       {
         // a bit of debug
         kodi::Log(ADDON_LOG_DEBUG, "session.initiate returns: sid=%s salt=%s", sid.c_str(), salt.c_str());
-
         std::string pinMD5 = kodi::GetMD5(m_settings.m_PIN);
         StringUtils::ToLower(pinMD5);
 
--- a/src/pvrclient-nextpvr.cpp
+++ b/src/pvrclient-nextpvr.cpp
@@ -580,13 +580,17 @@ bool cPVRClientNextPVR::OpenLiveStream(const kodi::addon::PVRChannel& channel)
 
 int cPVRClientNextPVR::ReadLiveStream(unsigned char* pBuffer, unsigned int iBufferSize)
 {
-  return m_livePlayer->Read(pBuffer, iBufferSize);
+  if (IsServerStreamingLive())
+  {
+    return m_livePlayer->Read(pBuffer, iBufferSize);
+  }
+  return -1;
 }
 
 void cPVRClientNextPVR::CloseLiveStream(void)
 {
   kodi::Log(ADDON_LOG_DEBUG, "CloseLiveStream");
-  if (m_livePlayer != nullptr)
+  if (IsServerStreamingLive())
   {
     m_livePlayer->Close();
     m_livePlayer = nullptr;
@@ -594,21 +598,23 @@ void cPVRClientNextPVR::CloseLiveStream(void)
   m_nowPlaying = NotPlaying;
 }
 
-
 int64_t cPVRClientNextPVR::SeekLiveStream(int64_t iPosition, int iWhence)
 {
-  int64_t retVal;
-  kodi::Log(ADDON_LOG_DEBUG, "calling seek(%lli %d)", iPosition, iWhence);
-  retVal = m_livePlayer->Seek(iPosition, iWhence);
-  kodi::Log(ADDON_LOG_DEBUG, "returned from seek()");
-  return retVal;
+  if (IsServerStreamingLive())
+  {
+    return m_livePlayer->Seek(iPosition, iWhence);
+  }
+  return -1;
 }
-
 
 int64_t cPVRClientNextPVR::LengthLiveStream(void)
 {
-  kodi::Log(ADDON_LOG_DEBUG, "seek length(%lli)", m_livePlayer->Length());
-  return m_livePlayer->Length();
+  if (IsServerStreamingLive())
+  {
+    kodi::Log(ADDON_LOG_DEBUG, "seek length(%lli)", m_livePlayer->Length());
+    return m_livePlayer->Length();
+  }
+  return -1;
 }
 
 PVR_ERROR cPVRClientNextPVR::GetSignalStatus(int channelUid, kodi::addon::PVRSignalStatus& signalStatus)
@@ -623,32 +629,39 @@ PVR_ERROR cPVRClientNextPVR::GetSignalStatus(int channelUid, kodi::addon::PVRSig
 
 bool cPVRClientNextPVR::CanPauseStream(void)
 {
-  if (m_nowPlaying == Recording)
+  if (IsServerStreaming())
   {
-    return true;
+    if (m_nowPlaying == Recording)
+      return true;
+    else
+      return m_livePlayer->CanPauseStream();
   }
-  else
-  {
-    bool retval = m_livePlayer->CanPauseStream();
-    return retval;
-  }
+  return false;
 }
 
 void cPVRClientNextPVR::PauseStream(bool bPaused)
 {
-  if (m_nowPlaying == Recording)
-    m_recordingBuffer->PauseStream(bPaused);
-  else
-    m_livePlayer->PauseStream(bPaused);
+  if (IsServerStreaming())
+  {
+    if (m_nowPlaying == Recording)
+      m_recordingBuffer->PauseStream(bPaused);
+    else
+      m_livePlayer->PauseStream(bPaused);
+  }
 }
 
 bool cPVRClientNextPVR::CanSeekStream(void)
 {
-  if (m_nowPlaying == Recording)
-    return true;
-  else
-    return m_livePlayer->CanSeekStream();
+  if (IsServerStreaming())
+  {
+    if (m_nowPlaying == Recording)
+      return true;
+    else
+      return m_livePlayer->CanSeekStream();
+  }
+  return false;
 }
+
 
 /************************************************************/
 /** Record stream handling */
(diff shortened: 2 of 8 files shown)


emveepee:
@sub @phunkyfish  could you review this please when you have time.  Thanks.  It would make my life easier if it goes in before PR 147 and I can focus on document the changes for Matrix.

src/pvrclient-nextpvr.cpp
```diff
@@ -665,69 +678,119 @@ bool cPVRClientNextPVR::OpenRecordedStream(const kodi::addon::PVRRecording& reco
...
+}
+
+bool cPVRClientNextPVR::IsServerStreamingRecording(bool log)
+{
+  if (m_nowPlaying == Recording && m_recordingBuffer->GetDuration() > 0)
+  {
+    return true;
+  }
+  if (log)
+    kodi::Log(ADDON_LOG_ERROR, "Unknown recording streaming state %d %d %d", m_nowPlaying, m_recordingBuffer->GetDuration(), m_livePlayer == nullptr);
```
phunkyfish: `!m_livePlayer`

pvr.nextpvr/resources/settings.xml
```diff
@@ -29,6 +29,14 @@
...
           </control>
         </setting>
+        <setting help="" id="hostprotocol" label="" type="string" option="hidden">
+          <level>4</level>
+          <default>http</default>
+          <control format="string" type="edit">
+            <heading></heading>
+            <hidden>true</hidden>
+          </control>
+        </setting>
```
phunkyfish, line 39: Can you explain this? I.e. why do you need a hidden setting?

phunkyfish (changes requested):

src/pvrclient-nextpvr.cpp
emveepee: I thought spelling out was clearer about what I wanted logged but I changed that.

pvr.nextpvr/resources/settings.xml
emveepee: I make it hidden because if it is a toggle the user might think they could simply enable https.  To make this work a user will need to setup  something like an nginix proxy (preferred) or manually configure the NextPVR server with a pfx file.  I figure most users will also use self signed certs for this and Kodi adds the complexity of requiring /system/certs/cacert.pem since I don't want to add a additional option to force override curl security.

phunkyfish: So if one of those is supplied you set this setting to store the value?

emveepee: No there would be no way to connect.  The user would need to edit the user settings.xml and this would need to be documented.  I did ask @sub3 to consider adding this for the future in the discovery API but I doubt https on the local LAN is important and discovery won't work over a WAN/Internet.

phunkyfish: Then I would add the documentation as part of this commit.

emveepee: I will be making a markdown file for the entire Matrix update which I have discussed with @Karellen  and this feature will be included but I certainly won't go into how to enable https just how to use https with a note about self-signed certs, but frankly this hidden setting will be one of the lowest priorities since there is no demand, I just didn't like the hard-coded http references in the code..  Do you want the commit removed until later?

phunkyfish: No, your plan sounds fine. I usually tend to add the documentation with commits. It useful for other contributors then, makes it easier to figure out what is going on.

phunkyfish (approved):

pvr.nextpvr/resources/settings.xml
emveepee: Thanks. I wish I didn't have to consider two communities (NextPVR and Kodi) with their separate forums and wikis.  No matter what or where I document I tend to believe if someone wants to use https they will still post "How do I ..." on one of the forums.

phunkyfish:
You need to check the cpp17 PR in kodi. The p8-platform patch will show you how to update StringUtils.

https://github.com/xbmc/xbmc/pull/18403

emveepee:
> 
> 
> You need to check the cpp17 PR in kodi. The p8-platform patch will show you how to update StringUtils.
> 
> [xbmc/xbmc#18403](https://github.com/xbmc/xbmc/pull/18403)

Thanks for the tip, I might have to cmake my setup from master I can't see where how even enable cpp17  I might lose these individual commits not worth cherry picking them

@phunkyfish  From a fresh build Windows has no problem building them (like the azure builds) perhaps the Windows builds still run cpp11.   Edit yes it does but I can force it to 17 and generate the same errors.  So if I fix the Windows builds I could break the other builds.

emveepee:
@AlwinEsch  the cpp17 error seems to be here https://github.com/xbmc/platform/blob/master/src/util/StringUtils.cpp#L456  in platform  I don't include my own String::Utils

pvr.hts is having the same issue.

phunkyfish:
> @AlwinEsch the cpp17 error seems to be here https://github.com/xbmc/platform/blob/master/src/util/StringUtils.cpp#L456 in platform I don't include my own String::Utils
> 
> pvr.hts is having the same issue.

You can add a patch to your depends. Changing p8-platform is problematic so I would just patch it.

Just add the patch to `depends/common/p8-platform`

All you need to fix this is the patch, nothing else. Your CI builds will be cpp-17 automatically.

pvr.nextpvr/resources/settings.xml
phunkyfish: Yes, it’s always tricky where to document it.

emveepee:
> 
> Just add the patch to `depends/common/p8-platform`

I tried that yesterday using the example for pvr.iptvsimple which patches zlib ok but with the patch file with the p8-platform.txt file it does not build.

I will try via the build system.  Edit:  Fails with the same error I get on Windows

2020-09-22T13:18:19.6357557Z   fatal: pathspec '.' did not match any files

phunkyfish:
Ok, but it works for non windows

phunkyfish:
So you need to exclude it for windows, just trying to remember how you do that.

emveepee:
> 
> 
> So you need to exclude it for windows, just trying to remember how you do that.

I don't think think is good because the only reason it works in Window without the patch is because is it isn't using c++17 As soon as I tell it to use c++17 the patch is required.

phunkyfish:
Correct. Once cpp17 is enabled for windows then that should be enabled too.

phunkyfish:
The patch is failing on windows for some reason. Not sure why

emveepee:
> 
> 
> The patch is failing on windows for some reason. Not sure why

"Performing gitconfig step: Disabling autocrlf to enable patching for 'p8-platform" is happening  before folders are created and the git download.  If I manually do those steps it works.

phunkyfish:
> > The patch is failing on windows for some reason. Not sure why
> 
> "Performing gitconfig step: Disabling autocrlf to enable patching for 'p8-platform" is happening before folders are created and the git download. If I manually do those steps it works.

Strange. I wonder how we fix that one. Any ideas @AlwinEsch ?

emveepee:
It will have something to do with git I changed to use a file as a test and it worked on windows, android and linux.  Not sure about those xcode errors they aren't related to StringUtils

phunkyfish:
Maybe try pushing again. The XCode errors may just be temp.

I guess once there is a workaround it can be used until we deprecate p8-platform fully.

emveepee:
> 
> Maybe try pushing again. The XCode errors may just be temp.

It feels more serious

```
Undefined symbols for architecture arm64:

_CVGetCurrentHostTime", referenced from:

      P8PLATFORM::CCondition<bool>::Wait(P8PLATFORM::CMutex&, bool&, unsigned int) in pvrclient-nextpvr.cpp.o

  "_CVGetHostClockFrequency", referenced from:

      P8PLATFORM::CCondition<bool>::Wait(P8PLATFORM::CMutex&, bool&, unsigned int) in pvrclient-nextpvr.cpp.o

ld: symbol(s) not found for architecture arm64
```

phunkyfish:
Hmmm, are there other changes since the previous release in there?

emveepee:
> 
> 
> Hmmm, are there other changes since the previous release in there?

I am assuming that is cpp17 on this line https://github.com/xbmc/platform/blob/master/src/util/timeutils.h#L81

phunkyfish:
That should I also work prior to cpp17 as far as I’m aware.

Let me give you a like to add to your cmake lists to see if it fixes it.

phunkyfish:
Try modifying to this in your CMakeLists.txt:

```
if(WIN32)
  list(APPEND DEPLIBS ws2_32)
  add_definitions(/D_WINSOCK_DEPRECATED_NO_WARNINGS)
elseif(CORE_SYSTEM_NAME STREQUAL osx OR
       CORE_SYSTEM_NAME STREQUAL darwin_embedded)
  find_library(COREVIDEO CoreVideo)

  include_directories(${COREVIDEO})

  list(APPEND DEPLIBS ${COREVIDEO})
endif()
```

I'm not entirely sure why this is required for cpp17 but I needed to add it to another addon. Worth a shot.

emveepee:
> 
> I'm not entirely sure why this is required for cpp17 but I needed to add it to another addon. Worth a shot.

Thank that did it and it is adding the patch so it should be cpp17.  I will try again later with the changes in platform.git, a new separate Matrix branch shouldn't add any impact on existing users.

phunkyfish:
I think this just shines a light on the fact that we need to remove the dependency on p8-platform from binary add-one as this is way too much work.

emveepee:
> 
> 
> I think this just shines a light on the fact that we need to remove the dependency on p8-platform from binary add-one as this is way too much work.

I think the opposite, it is less work especially for addons that aren't heavily supported.  If every binary addon had a custom StringUtils each developer would have to figure out how patch their own copy rather then just change a git hash to known value.  Updating the hash could/should  have been part of the cpp 17 PR with a full build of supported binary addon like @AlwinEsch  does would have place the work on the PR where it should be.

Moving to std::thread is not too difficult but it will be a bit of time and testing.

phunkyfish:
That’s still removing the dependency. I agree it should be supported directly by the addon interface.

emveepee:
@phunkyfish @AlwinEsch Is there any change in using this patch https://github.com/emveepee/platform/commit/e7e0a05f05f01db29868e568ea233b7bc963fb0d so I can get this PR out of limbo?   Using PR148 as a test it passes TI https://travis-ci.org/github/kodi-pvr/pvr.nextpvr better than current patches.

phunkyfish:
The cpp17 PR will be reverted temporally to allow the alpha2 release to be tagged and built. I recommend you get your PR through then as p8 will build as it did normally. After that the cpp17 PR will go back but there will be StringUtils, and some thread helper classes to help with p8 removal so not so bad if you miss that window.

emveepee:
> 
> 
> The cpp17 PR will be reverted temporally to allow the alpha2 release to be tagged and built. I recommend you get your PR 

Thanks I reverted to the approved PR  from  a couple of week ago.   @sub3 could you merge this but not tag?  PR 149 for cast and crew  is pending this PR but the comments on this PR are confusing enough to warrant a separate PR.   Hopefully we can merge and tag PR 149 after PR 18403 is merged in Kodi.

AlwinEsch:
I will also try to remove the p8-platform from it (not looked now, hope not to much used from them).

emveepee:
> 
> 
> I will also try to remove the p8-platform from it (not looked now, hope not to much used from them).

@AlwinEsch  can't we we wait for post alpha2 for this?   I'd like to get this and PR149 in user hands soon, the addon has been blocked for a while,  and then I can start testing and adding std::threads.  If you can focus on providing a good StringUtils  for all platforms for cpp17 that would be most helpful.

AlwinEsch:
NP, think to have also not so soon done :grin: 

About StringUtils is already in Kodi. With Threads is also now a request present https://github.com/xbmc/xbmc/pull/18507 and the std::threads not direct needed.

Saw at pvr.argustv that it would be damn difficult to have the std::thread usable without helper classes, so I created more helper classes in addon headers for it.

emveepee:
> 
> About StringUtils is already in Kodi. 
>

@AlwinEsch is the new StringUtils supposed to be generating more warnings then before?  It is much harder to review my Windows self builds and the azure build logs for warnings now because they are filled with repeated warning from StringUtils.h that I don't remember seeing before.  It isn't just pvr.nextpvr I checked a recent build of pvr.vbox and saw the  same thing https://dev.azure.com/teamkodi/219b1c81-3a34-4614-8eac-b6a2ecee08fb/_apis/build/builds/2483/logs/31

AlwinEsch:
Will look to fix the warnings, I didn't notice them before under Linux. 

These are probably also in Kodi itself with its StringUtils.cpp, only it comes there only once.

PR #147, opened 2020-09-13
[Matrix] increase version to 8.0.0 (API related)

AlwinEsch:
API related to xbmc/xbmc#17438 (DEMUX_PACKET not used there)

### DIFF
--- a/pvr.nextpvr/addon.xml.in
+++ b/pvr.nextpvr/addon.xml.in
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <addon
   id="pvr.nextpvr"
-  version="7.1.1"
+  version="8.0.0"
   name="NextPVR PVR Client"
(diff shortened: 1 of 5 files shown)


emveepee:
Can you update PR 137 to remove this?  I have been holding back on changing the other issues you identified in PR 137 and if you remove it you could update the change log and merge PR 137 right away.

AlwinEsch:
Can swap them over and make for inputstream with here.

emveepee:
@AlwinEsch I assume this needs to be merged now for STREAM_TIME_BASE

AlwinEsch:
There is still the perverse stupid launchpad problem, on which Kodi's build is only triggered one time per day and if addon comes in before, does his build fail :unamused:.

For this we must wait mostly 24 hours before there ready.

Looking currently to find a way to prevent this, see "addon-system-external" on Slack.

PR #148, opened 2020-09-23
Use updated c++17 platform 

emveepee:
Test of changed platform to avoid hacking pvr.nextpvr for c++17 changes

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,16 +1,12 @@
 language: cpp
 
-#
-# Define the build matrix
-#
-# Travis defaults to building on Ubuntu Trusty when building on
-# Linux. We need Xenial in order to get up to date versions of
-# cmake and g++.
-#
 env:
   global:
     - app_id=pvr.nextpvr
 
+#
+# Define the build matrix
+#
 matrix:
   include:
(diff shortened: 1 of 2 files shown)


emveepee:
@phunkyfish this is how I believe it should be down pointing to platform with the change https://github.com/emveepee/platform/commit/4016b39b3ab1709f60befc1e9b73725f9471e62d

phunkyfish:
I agree but changing platform is not so simple. As git is used in shallow mode to pull the revision that will disappear if we add a commit.

As it is called so many times github will throttle and it will be the source of many issues. Remember it is shared across kodi 18 and 19.

phunkyfish:
I did the last time and broke lots of stuff. It’s how I found out not to do that!

emveepee:
> 
> 
> I did the last time and broke lots of stuff. It’s how I found out not to do that!

If a submit a PR to new branch not to default/master isn't the only access via the hash?  I would just need a new branch to submit to.  I checked and there are other branches on it so branches are possible without conflict.

emveepee:
This is closed as alternate solutions are being worked on

PR #149, opened 2020-10-03
Add cast and crew

emveepee:
Versions 5.0.9 of the NextPVR backend includes cast and crew.  This adds xml support for it.  Older versions of NextPVR will not have the markup so there are no compatibility concerns.

### DIFF
--- a/pvr.nextpvr/addon.xml.in
+++ b/pvr.nextpvr/addon.xml.in
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <addon
   id="pvr.nextpvr"
-  version="7.0.1"
+  version="7.0.2"
   name="NextPVR PVR Client"
   provider-name="Graeme Blackley">
   <requires>@ADDON_DEPENDS@
--- a/pvr.nextpvr/changelog.txt
+++ b/pvr.nextpvr/changelog.txt
@@ -1,3 +1,6 @@
+v7.0.2
+- Add cast and crew from NextPVR 5.0.9
+
 v7.0.1
 - Protect against out media calls from Kodi on closed files
 - Formalize and test manual https support
--- a/pvr.nextpvr/changelog.txt
+++ b/pvr.nextpvr/changelog.txt
@@ -1,5 +1,6 @@
 v7.0.2
 - Add cast and crew from NextPVR 5.0.9
+- Fix regression in discovery
 
 v7.0.1
 - Protect against out media calls from Kodi on closed files
--- a/pvr.nextpvr/resources/language/resource.language.en_gb/strings.po
+++ b/pvr.nextpvr/resources/language/resource.language.en_gb/strings.po
@@ -313,3 +313,7 @@ msgstr ""
 msgctxt "#30196"
 msgid "Open NextPVR settings"
 msgstr ""
+
+msgctxt "#30197"
+msgid "Show guide cast, director and writer"
+msgstr ""
--- a/pvr.nextpvr/resources/settings.xml
+++ b/pvr.nextpvr/resources/settings.xml
@@ -257,6 +257,16 @@
             </dependency>
           </dependencies>
         </setting>
+        <setting help="" id="castcrew" label="30197" type="boolean">
+          <level>1</level>
+          <default>false</default>
+          <control type="toggle"/>
+          <dependencies>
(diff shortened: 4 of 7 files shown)


phunkyfish (approved):
LGTM

phunkyfish:
If you push this again now it should build fully.

emveepee:
> 
> 
> If you push this again now it should build fully.

Ok thanks, when I add the change log etc that will work.  I expect I will need enhance this once I see user data.  It expect is too centric to my English data source.

I also have concern how many people will even use this given the Kodi Estuary implementation https://i.imgur.com/YhI911t.mp4  I am receiving comments before it is released here https://forum.kodi.tv/showthread.php?tid=357311&pid=2979782#pid2979782 so the default will probably be changed to false.

It is perhaps most unfortunate that PVR recording don't include cast and crew where I expect I would use it most.  I'd probably update my own recordings with the actors' role but can't see doing that on every guide item.

emveepee:
@phunkyfish  could you please review the second commit,  The alpha2 release came as quite  of a surprise after two weeks of limbo and I found a regression, when I did my pre-release testing.  I hope in the future there is an attemp to limit breaking API bumps to match the alpha since I find very few users upgrade to nightlies to help with the testing.   By this I suggest API 8 should be released near to alpha3 or 4.

@sub3 when this is approved could you please merge and tag.

phunkyfish (approved):

phunkyfish:
It's the nature of Alpha's that you should not install it unless you are willing to accept you might need a nightly build to fix things.

That being said we do try to minimise the issues caused the API changes but I'm not seeing what the cause of this was. We have not bumped PVR API in a while.

emveepee:
> 
> 
> It's the nature of Alpha's that you should not install it unless you are willing to accept you might need a nightly build to fix things.
> 

I fully understand but I think there is a user perception that an alpha is more thoroughly tested than a nightly and more than a milestone.

> That being said we do try to minimise the issues caused the API changes but I'm not seeing what the cause of this was. We have not bumped PVR API in a while.

It wasn't a PVR API, on PR 18408 it was noted "A lot of addons need version increase" so when it was released  pvr.nextpvr alpha users starting getting confusing prompts to update in the alpha that failed because of the missing dependency.

phunkyfish:
Ok, this was a change I was not aware either. Apologies for that.

Regarding the testing that will only be on kodi but not across PVR add-ons other than the ones the devs use. Really the perception should be that a beta is more reliable. An Alpha is really a point release of a nightly. It lets the community know we are getting ready for a beta but it will most likely not work in many parts.

AlwinEsch:
I don't like it directly either, just the addon system has to be through before the beta comes, even if it's hell.

Therefore, the issue of version increase has to be addressed again, I think increasing the major with every API change is senseless. Since there will soon be a gingatic version difference between Leia and Matrix, nothing different about the addon itself.

Unfortunately there will be at least two more changes.

What's to come:
- Convert input stream + video codec
   - This in smaller steps
- Finally increase the PVR API with inputstream as it unfortunately uses inputstream parts
- Then as last the attribute hidden problem which is necessary for a secure API
   - Without, a class of addon "A" can accidentally end up in class of addon "B" (if both with same name)
   - There need every addon a increase

emveepee:
> 
> Regarding the testing that will only be on kodi but not across PVR add-ons other than the ones the devs use. Really the perception should be that a beta is more reliable. An Alpha is really a point release of a nightly. It lets the community know we are getting ready for a beta but it will most likely not work in many parts.

I myself was mistaken and pretty much leaft my Android devices on the alpha because of hassles getting the apk.   I am also spoiled by the stability of PVR which has worked very well in Matrix and it appears better than Leia.

emveepee:
> 
> 
> I don't like it directly either, just the addon system has to be through before the beta comes, even if it's hell.

I understand that this is a necessary evil and not at all meant as a complaint.  If was just suggesting that there might be a way to make the breaking API changes come closer to an alpha milestone so the users aren't prompted to update or miss changes for too long.

I was also wrong in my interpretation of the alpha release cycle stability since I was suggesting to users that they stay on the alpha when they probably should migrate.

phunkyfish:
An alpha on platforms that bundle binary addons can be stable. But Android, windows and mac would not be as they will auto update add-ons.

Really the most stable platform is Leia!

PR #150, opened 2020-10-06
Remove p8-platform

emveepee:
Remove dependencies on p8 platform

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,11 +4,9 @@ project(pvr.nextpvr)
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${PROJECT_SOURCE_DIR})
 
 find_package(Kodi REQUIRED)
-find_package(p8-platform REQUIRED)
 find_package(TinyXML2 REQUIRED)
 
-include_directories(${p8-platform_INCLUDE_DIRS}
-                    ${TINYXML2_INCLUDE_DIR}
+include_directories(${TINYXML2_INCLUDE_DIR}
                     ${KODI_INCLUDE_DIR}/..) # Hack way with "/..", need bigger Kodi cmake rework to match right include ways
 
 set(NEXTPVR_SOURCES src/addon.cpp
@@ -55,11 +53,10 @@ set(NEXTPVR_HEADERS src/addon.h
                     src/buffers/Seeker.h
                     src/utilities/XMLUtils.h)
 
-SET(DEPLIBS ${p8-platform_LIBRARIES}
-            ${TINYXML2_LIBRARIES})
+SET(DEPLIBS ${TINYXML2_LIBRARIES})
 if(WIN32)
   list(APPEND DEPLIBS ws2_32)
-  add_definitions(/D_WINSOCK_DEPRECATED_NO_WARNINGS)
+  add_definitions(-D_WINSOCKAPI_ -D_WINSOCK_DEPRECATED_NO_WARNINGS)
 endif()
 
 build_addon(pvr.nextpvr NEXTPVR DEPLIBS)
--- a/src/pvrclient-nextpvr.h
+++ b/src/pvrclient-nextpvr.h
@@ -28,7 +28,6 @@
 #include "buffers/RollingFile.h"
 #include "buffers/TimeshiftBuffer.h"
 #include "buffers/TranscodedBuffer.h"
-#include "p8-platform/threads/threads.h"
 #include <map>
 
 enum eNowPlaying
@@ -40,7 +39,7 @@ enum eNowPlaying
   Transcoding
 };
 
-class ATTRIBUTE_HIDDEN cPVRClientNextPVR : public kodi::addon::CInstancePVRClient, P8PLATFORM::CThread
+class ATTRIBUTE_HIDDEN cPVRClientNextPVR : public kodi::addon::CInstancePVRClient
 {
 public:
   /* Class interface */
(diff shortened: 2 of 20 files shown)


phunkyfish:
You can also have a `using namespace kodi::tools;` if you don't want to namespace each call to StringUtils.

emveepee:
> 
> 
> You can also have a `using namespace kodi::tools;` if you don't want to namespace each call to StringUtils.

I will have a look at that this, because it is new this is clearer to me right now.  I find variable m_threadstop a bit confusing without any qualifier and I don't like things too hidden.

phunkyfish:
There is no need to use the thread class in kodi tools. std::thread should be just fine.

That class was added for cases that are too dependent on p8 threading.

phunkyfish:
Example: https://github.com/kodi-pvr/pvr.dvbviewer/pull/100/commits/d039b29429b189e0395bc917f5f7cc9b707c0af3

emveepee:
@AlwinEsch  kodi.binary.global.tools is not part of of @ADDON_DEPENDS@ so should I specify it in addon.xml.in ? I am hoping that this build will released after my testing as part of alpha2 so I'd like to keep the minversion at 1.0.3 if possible if that is a concern.

AlwinEsch:
Ah no, was forgotten to set in version.h will add to https://github.com/xbmc/xbmc/pull/18530

emveepee:
> 
> 
> Ah no, was forgotten to set in version.h will add to [xbmc/xbmc#18530](https://github.com/xbmc/xbmc/pull/18530)

Ok that is clearer that it is a full version bump.  I still have two questions.  Do I need to specify that I may only need 1.0.3 for pvr.nextpvr since it appears to compile and run fine even though we do have a winsock definition.

  add_definitions(/D_WINSOCK_DEPRECATED_NO_WARNINGS)

Also will there be any issues running compiles made with 1.0.4 with alpha2 which is at 1.0.3

emveepee:
@phunkyfish and @AlwinEsch if you approve this change and it doesn't impact alpha2 feel free to merge and tag it.  I can combine the two threads commits into one if you feel it would be better, I did this as a check to see if I was missing something.

AlwinEsch:
Maybe cleanup the commits a bit as no. 3 remove no. 2.

emveepee:
Well that answers my question about the winsock change.   Back to rebuild my dev environment again.

src/pvrclient-nextpvr.h
```diff
@@ -28,7 +28,7 @@
 #include "buffers/RollingFile.h"
 #include "buffers/TimeshiftBuffer.h"
 #include "buffers/TranscodedBuffer.h"
-#include "p8-platform/threads/threads.h"
+#include "kodi/tools/Thread.h"
```
phunkyfish: No need to include this. Instead include `atomic` and `thread` standard headers

phunkyfish (changes requested):

CMakeLists.txt
```diff
@@ -55,11 +53,11 @@ set(NEXTPVR_HEADERS src/addon.h
                     src/buffers/Seeker.h
                     src/utilities/XMLUtils.h)
 
-SET(DEPLIBS ${p8-platform_LIBRARIES}
-            ${TINYXML2_LIBRARIES})
+SET(DEPLIBS ${TINYXML2_LIBRARIES})
 if(WIN32)
   list(APPEND DEPLIBS ws2_32)
   add_definitions(/D_WINSOCK_DEPRECATED_NO_WARNINGS)
+  add_definitions(-D _WINSOCKAPI_)
```
AlwinEsch: You was faster 😄.

Thought to write:
```cmake
  add_definitions(-D_WINSOCKAPI_ -D_WINSOCK_DEPRECATED_NO_WARNINGS)
```

Think as one line looks better and also is the `/` not direct needed.

**EDIT** Also without space as I thing not set always correct.

src/pvrclient-nextpvr.h
emveepee: Thanks I missed removing that.  atomic and thread are already in the general Buffer class.

phunkyfish: Edit, apologies. The atomic and thread did not appear in my original post.

CMakeLists.txt
emveepee: Yes thanks that looks much cleaner

phunkyfish (approved):

phunkyfish:
@AlwinEsch let me know if you are happy and I will merge this.

emveepee:
Its nice to see Travis integrated.  Do you know what this means?

```
dpkg-shlibdeps: warning: debian/kodi-pvr-nextpvr/usr/lib/x86_64-linux-gnu/kodi/addons/pvr.nextpvr/pvr.nextpvr.so.7.1.0 contains an unresolvable reference to symbol pthread_create: it's probably a plugin
```

AlwinEsch:
> Its nice to see Travis integrated. Do you know what this means?
> 
> ```
> dpkg-shlibdeps: warning: debian/kodi-pvr-nextpvr/usr/lib/x86_64-linux-gnu/kodi/addons/pvr.nextpvr/pvr.nextpvr.so.7.1.0 contains an unresolvable reference to symbol pthread_create: it's probably a plugin
> ```

No idea about it, tested also at home and there not come :thinking:

AlwinEsch (approved):
Looks good to me.

@emveepee think the runtime test was OK about?

emveepee:
> 
> @emveepee think the runtime test was OK about?

Runtime test was fine but I can't say I checked all the StringUtils calls.  It worked with alpha2 as well with tools 1.0.3 

> 
 ```
dpkg-shlibdeps: warning: debian/kodi-pvr-nextpvr/usr/lib/x86_64-linux-gnu/kodi/addons/pvr.nextpvr/pvr.nextpvr.so.7.1.0 contains an unresolvable reference to symbol pthread_create: it's probably a plugin
```
> 
> No idea about it, tested also at home and there not come 🤔

mediaportal has it in Travis as well when   env: DEBIAN_BUILD=true  so since it is not only pvr.nextpvr I guess we can ignore it for now.

PR #153, opened 2020-10-13
Show drive space usage

emveepee:
Optionally show drive space by default drive or by summing all drives

### DIFF
--- a/src/Recordings.cpp
+++ b/src/Recordings.cpp
@@ -13,6 +13,7 @@
 #include "pvrclient-nextpvr.h"
 
 #include <regex>
+#include <unordered_set>
 
 #include <kodi/tools/StringUtils.h>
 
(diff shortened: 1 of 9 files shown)


src/Recordings.cpp
```diff
@@ -119,6 +177,11 @@ PVR_ERROR Recordings::GetRecordings(bool deleted, kodi::addon::PVRRecordingsResu
       }
     }
     m_iRecordingCount = recordingCount;
+    // force read disk space
+    m_checkedSpace = 0;
+    uint64_t total;
+    uint64_t used;
+    GetDriveSpace(total, used);
```
phunkyfish, line 184: Why have this here?

emveepee: > 
> Why have this here?

The check to the backend is every 5 minutes.  This recording list update is triggered right after a user does maintenance so by forcing the update, the user will get dynamic updates as shows are removed and timers scheduled to provide this information in a timely manner.

phunkyfish: Cool, makes sense.

phunkyfish (approved):
LGTM

PR #155, opened 2020-11-02
Standardize backend requests

emveepee:
Move XML validation to single location for consistent error checking on backend requests.   Remove static strings from function calls for readability.

### DIFF
--- a/src/pvrclient-nextpvr.cpp
+++ b/src/pvrclient-nextpvr.cpp
@@ -207,6 +207,14 @@ ADDON_STATUS cPVRClientNextPVR::Connect(bool sendWOL)
   return status;
 }
 
+
+void cPVRClientNextPVR::ResetConnection()
+{
+  m_nextServerCheck = 0;
+  m_connectionState = PVR_CONNECTION_STATE_DISCONNECTED;
+  m_bConnected = false;
+}
+
 void cPVRClientNextPVR::Disconnect()
 {
   m_request.DoActionRequest("session.logout");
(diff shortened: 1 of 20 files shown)


emveepee:
The first commit has a lot of change because one level of if indenting has been removed in much of the code by moving xml processing to the calling function and returning and XML document rather then strings.  

@phunkyfish  is there any advantage to changing these calls to pass doc by reference?  Currently they follow this template

```
  tinyxml2::XMLDocument doc;
  if (m_request.DoMethodRequest("recording.list&filter=all", doc) == tinyxml2::XML_SUCCESS)
```

phunkyfish:
Yes, always pass by reference where possible as it will avoid unnecessary copying.

src/pvrclient-nextpvr.cpp
```diff
@@ -660,8 +660,9 @@ bool cPVRClientNextPVR::CanSeekStream(void)
 {
   if (IsServerStreaming())
   {
+    //  function is not called for recordings
```
phunkyfish: The comment conflicts with the code here. Maybe expand so it is clear.

emveepee: The documentation is not clear on how recordings deal with this.  Perhaps the best solution for now is to remove recording logic from that function entirely and assume it applies only to live tv to be consistent with other addons.

phunkyfish: Can seek stream is not for recordings at all. At least I never thought it was.

emveepee: Semantics I guess.  Here https://codedocs.xyz/AlwinEsch/kodi/group__cpp__kodi__addon__pvr___streams.html identifies 3 stream types.  If the function only applies to TV is probably should be under TV.

phunkyfish: Agreed the docs are not clear at all on that. Could you PR an update so we clean that up for matrix?

I think being specific even where it is currently and just specifying live only would be ok.

emveepee: NP I will have a look.

phunkyfish (approved):
LGTM

PR #157, opened 2020-12-16
Rework recording watched state

emveepee:
Maintain play count to manage watched state.  Handle fake return position 0 on resume.  Allow option Kodi for management of resume state.

### DIFF
--- a/pvr.nextpvr/addon.xml.in
+++ b/pvr.nextpvr/addon.xml.in
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <addon
   id="pvr.nextpvr"
-  version="8.0.2"
+  version="8.0.3"
   name="NextPVR PVR Client"
   provider-name="Graeme Blackley">
   <requires>@ADDON_DEPENDS@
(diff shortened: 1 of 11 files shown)


emveepee:
@phunkyfish  here are my changes with a stored fake play count.  I added my findings on SetRecordingsLastPlayPostion() and SetRecordingPlayCount() in case others wonder when Kodi send 0 and -1 as the lastplayedposition.  I decided to allow an option to ignore the backend and use the database fields to allow by client settings if desired.

phunkyfish (approved):
Nice

emveepee:
@phunkyfish  I added another commit to deal with crashing issue with EPG based timer.  I thought there might be a way to resolvr this in Kodi  but @ksooo  advised that my solution wouldn't work and there no other current solution.  These value are not shown to user since they are EPG based but they can't be zero because of negative time_t issues.

phunkyfish (approved):
Cool, looks fine to me.

emveepee:
@phunkyfish resending this for review of another commit.  Sorry I had stashed this latest commit and forgot about it for a couple of weeks.  I've been using it and thought it was already submitted.

phunkyfish (approved):

PR #159, opened 2021-01-08
Fix std::chrono::duration_cast

emveepee:
Need to use count() to get the value.

### DIFF
--- a/pvr.nextpvr/addon.xml.in
+++ b/pvr.nextpvr/addon.xml.in
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <addon
   id="pvr.nextpvr"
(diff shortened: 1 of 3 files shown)


phunkyfish (approved):

phunkyfish:
Want to increase the version 8.1.0 and add the PVR API changelog note from here: https://github.com/kodi-pvr/pvr.nextpvr/pull/158

Then I can just merge this once the PVR API change is merged.

emveepee:
> 
> 
> Want to increase the version 8.1.0 and add the PVR API changelog note from here: #158
 
Sure there is the issue of changing int64_t to int when I use the count() in this change.  I wanted to see some warnings and I will be adding the  static_cast<int> too

PR #160, opened 2021-01-09
Correct manual timer

emveepee:
Allow modify of recording duration when user doesn't enter dialog fields correctly.

### DIFF
--- a/pvr.nextpvr/resources/language/resource.language.en_gb/strings.po
+++ b/pvr.nextpvr/resources/language/resource.language.en_gb/strings.po
@@ -303,7 +303,7 @@ msgid "Use backend genre text"
 msgstr ""
 
 msgctxt "#30694"
-msgid "Kodi default is DVB genre text.  Default required for Kodi genre search and language translation"
+msgid "Kodi default is DVB genre text. Default required for Kodi genre search and language translation"
 msgstr ""
 
 msgctxt "#30195"
@@ -340,4 +340,8 @@ msgstr ""
 
 msgctxt "#30700"
 msgid "When disabled resume location and watched status will be managed only in Kodi"
+msgstr ""
+
+msgctxt "#30201"
+msgid "End time invalid. Choose a recording duration"
 msgstr ""
--- a/src/Timers.cpp
+++ b/src/Timers.cpp
@@ -12,6 +12,7 @@
 #include "pvrclient-nextpvr.h"
 #include <kodi/General.h>
 #include <kodi/tools/StringUtils.h>
+#include <kodi/gui/dialogs/Select.h>
 #include <string>
 
 using namespace NextPVR;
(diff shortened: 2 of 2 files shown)


src/Timers.cpp
```diff
@@ -616,7 +640,7 @@ PVR_ERROR Timers::AddTimer(const kodi::addon::PVRTimer& timer)
       timer.GetClientIndex(),
       timer.GetClientChannelUid(),
       (int)timer.GetStartTime(),
-      (int)(timer.GetEndTime() - timer.GetStartTime()),
+      (int)(endTime - timer.GetStartTime()),
```
phunkyfish: C++ cast please

Might as well fix the line before this one while you are at it ;)

src/Timers.cpp
```diff
@@ -606,6 +607,29 @@ PVR_ERROR Timers::AddTimer(const kodi::addon::PVRTimer& timer)
...
+  {
+    std::vector<std::string> options;
+    int offset = 0;
+    for (unsigned int minutes = 30; minutes <= 180; minutes += 30)
+    {
+      const std::string option = kodi::tools::StringUtils::Format("%d %s", minutes, kodi::GetLocalizedString(12391).c_str());
+      options.emplace_back(option);
+    }
+    int option = kodi::gui::dialogs::Select::Show(kodi::GetLocalizedString(30201), options, 1);
+    if (option > 0)
```
phunkyfish: Should this not be >= 0?

-1 is returned if nothing is selected.

pvr.nextpvr/resources/language/resource.language.en_gb/strings.po
```diff
@@ -341,3 +341,7 @@ msgstr ""
 msgctxt "#30700"
 msgid "When disabled resume location and watched status will be managed only in Kodi"
 msgstr ""
+
+msgctxt "#30201"
+msgid "End time invalid.  Choose a recording duration"
```
phunkyfish: One space between the sentences.

phunkyfish (changes requested):
Just some minors and a possible functional error.

src/Timers.cpp
emveepee: There were a lot so I was trying to be consistent so I guess I must change them all.  Done

src/Timers.cpp
emveepee: Good catch @phunkyfish thanks.  I was going back and fourth between different options, number entry, time entry, so missed this.  It is too bad there is no slider dialog in the GUI.

pvr.nextpvr/resources/language/resource.language.en_gb/strings.po
emveepee: I'm an old guy so learned to type on a typewriter and this is automatic.  Fixed.

ksooo:
Shouldn't this be fixed in Kodi core?

phunkyfish:
> Shouldn't this be fixed in Kodi core?

Sure, I had a look but couldn't easily figure out how it should be done. The timer dialog code is not the easiest to follow. The code also doesn't deal well with different error responses. 

Firstly the timer dialog should be smarter and it should also allow for an invalid parameters response from the backend when a timer is added. Currently it just says check the log.

emveepee:
> 
> Shouldn't this be fixed in Kodi core?

There is no argument there but as I posted on the forum ideally it requires duration not start and stop times or users may still end up forgetting to change the end time.  The dialog also allows dates in the past, > 24h recordings etc.  which seemed bigger than an rc2 level change.  

If it was implemented now I see it like the recurring recording start and end times, adding a start date.  

Note that users have asked for EPG based recordings to have a start date to allow future recordings but again that is likely for v20.

ksooo:
https://github.com/xbmc/xbmc/blob/master/xbmc/pvr/dialogs/GUIDialogPVRTimerSettings.cpp#L590

@phunkyfish

phunkyfish:
> https://github.com/xbmc/xbmc/blob/master/xbmc/pvr/dialogs/GUIDialogPVRTimerSettings.cpp#L590
> 
> @phunkyfish

Yes, I found the same line. But I couldn’t see how to a) cancel the Save and display and error message or b) make the dialog more clever by setting a better end time once a start time is set.

ksooo:
@phunkyfish overriding `CGUIDialogSettingsManualBase::OnSettingChanging` might do the trick.

phunkyfish:
Cool. I’ll take a look.

In the meantime I added a sensible default so at the least the timer is valid: https://github.com/xbmc/xbmc/pull/19050

emveepee:
PR https://github.com/xbmc/xbmc/pull/19050 removes the need for this PR.  Thanks @phunkyfish  Closed.

PR #161, opened 2021-01-14
Improve startup behaviour and SID renewal, timer changes,

emveepee:
Multiple commits for final release.  Flag invalid channels, can toggle timer state, renew authentication during playback, remove requirement to cache the event uid, improved triggering for EPG updates.

### DIFF
--- a/src/pvrclient-nextpvr.cpp
+++ b/src/pvrclient-nextpvr.cpp
@@ -306,9 +306,21 @@ bool cPVRClientNextPVR::IsUp()
             {
               if (lastUpdate > m_lastEPGUpdateTime)
               {
-                SetConnectionState("Force update", PVR_CONNECTION_STATE_UNKNOWN);
-                SetConnectionState("Reload from backend", PVR_CONNECTION_STATE_CONNECTED);
+                // trigger EPG updates for all channels with a guide source
+                kodi::Log(ADDON_LOG_DEBUG, "Trigger EPG update start");
+                int channels = 0;
+                for (const auto &updateChannel : m_channels.m_channelDetails)
+                {
+                  if (updateChannel.second.first == false)
+                  {
+                    channels++;
+                    TriggerEpgUpdate(updateChannel.first);
+                  }
+                }
+                kodi::Log(ADDON_LOG_DEBUG, "Triggered %d channel updates", channels);
+
                 m_lastEPGUpdateTime = lastUpdate;
+                m_lastRecordingUpdateTime = update_time;
                 return m_bConnected;
               }
             }
--- a/src/pvrclient-nextpvr.cpp
+++ b/src/pvrclient-nextpvr.cpp
@@ -357,13 +357,17 @@ bool cPVRClientNextPVR::IsUp()
         }
       }
     }
-    else if (m_nowPlaying == Transcoding)
+    else if (m_nowPlaying != NotPlaying)
     {
-      if (m_livePlayer->IsRealTimeStream() == false)
+      m_request.RenewSID();
+      if (m_nowPlaying == Transcoding)
       {
-        //m_livePlayer->Close();
-        m_nowPlaying = NotPlaying;
-        m_livePlayer = nullptr;
+        if (m_livePlayer->IsRealTimeStream() == false)
+        {
+          //m_livePlayer->Close();
+          m_nowPlaying = NotPlaying;
+          m_livePlayer = nullptr;
+        }
       }
     }
   }
--- a/src/pvrclient-nextpvr.cpp
+++ b/src/pvrclient-nextpvr.cpp
@@ -113,7 +113,6 @@ cPVRClientNextPVR::~cPVRClientNextPVR()
   delete m_timeshiftBuffer;
   delete m_recordingBuffer;
   delete m_realTimeBuffer;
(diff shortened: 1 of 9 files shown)


emveepee:
@phunkyfish @sub3 can I have this reviewed and hopefully tagged and merged.  The second commit is pretty important.

I was surprised by the rc1 today, I had hoped to get this fix in for AppleTV users it is important, I guess I was expecting the Kodi team would wait for the Window dav1d build to be in sync with other platforms before going to rc.

phunkyfish:
I was not aware the windows dav1d build was different for windows.

emveepee:
> 
> 
> I was not aware the windows dav1d build was different for windows.

Yes I watch PR https://github.com/xbmc/xbmc/pull/19016 because it requires a different download step as noted

phunkyfish (approved):

phunkyfish:
If you have something lined up for next RC please let me know so we can make sure it's in on time.

emveepee:
> 
> 
> If you have something lined up for next RC please let me know so we can make sure it's in on time.

Thanks.  There will be one after I can start testing a new backend API that supports enabling/disabling timers but that won't take any time.

src/pvrclient-nextpvr.cpp
```diff
@@ -357,13 +357,17 @@ bool cPVRClientNextPVR::IsUp()
         }
       }
     }
-    else if (m_nowPlaying == Transcoding)
+    else if (m_nowPlaying != NotPlaying)
```
phunkyfish, line 371: Is `NotPlaying` a constant/enum?

emveepee: > 
> 
> Is `NotPlaying` a constant/enum?

It's an enum 

```
enum eNowPlaying
{
  NotPlaying = 0,
  TV = 1,
  Radio = 2,
  Recording = 3,
  Transcoding
};
```

phunkyfish: Cool, I think I’m just used to enums being CAPS_AND_UNDERSCORES so it looked a little odd to me.

emveepee:
@phunkyfish here are two changes that deal with last months core change.  There is no need to cache the backend EPG oid since the addon get it on demand.  It is a light call to the backend well worth doing and it will be used infrequently.

The other change triggers EPG updates properly.  I only tested with 300 channels but those were handled easily.

All in all, as your wrote on Slack the core PR combined with these changes will bring very good performance changes for the user, even if it did come late in the development cycle.

I have the other change I posted about ready too, do you know if there is a way to have the timer enabled  toggle not appear on the create screen?  The backend doesn't support adding a disabled timer.

phunkyfish:
The timer type specifies the usage I believe.

`PVR_TIMER_TYPE_SUPPORTS_ENABLE_DISABLE`

emveepee:
> 
> 
> The timer type specifies the usage I believe.
> 
> `PVR_TIMER_TYPE_SUPPORTS_ENABLE_DISABLE`

That part is working fine, but I was wondering if there was a why to have it not provide the choice when creating a new timer.

phunkyfish:
> > The timer type specifies the usage I believe.
> > `PVR_TIMER_TYPE_SUPPORTS_ENABLE_DISABLE`
> 
> That part is working fine, but I was wondering if there was a why to have it not provide the choice when creating a new timer.

Does it not hide/show depending on the chosen time type?

src/pvrclient-nextpvr.cpp
```diff
@@ -306,9 +306,21 @@ bool cPVRClientNextPVR::IsUp()
             {
               if (lastUpdate > m_lastEPGUpdateTime)
               {
-                SetConnectionState("Force update", PVR_CONNECTION_STATE_UNKNOWN);
-                SetConnectionState("Reload from backend", PVR_CONNECTION_STATE_CONNECTED);
+                // trigger EPG updates for all channels with a guide source
+                kodi::Log(ADDON_LOG_DEBUG, "Trigger EPG update start");
+                int channels = 0;
+                for ( const auto &updateChannel : m_channels.m_channelDetails )
```
phunkyfish: `for (const auto &updateChannel : m_channels.m_channelDetails)`

phunkyfish (approved):
Just the one cosmetic minor

emveepee:
> 
> Does it not hide/show depending on the chosen time type?

Everything is fine.  I just don't want it displayed on create.   Adding new type with it off just makes the list bigger and still gives this option, I was hoping some sort of tag would help.  I probably just need to live with it and am too picky.  It is like the menuhook, it gives Activate/Deactivate the Dialog toggle says Enable.

src/pvrclient-nextpvr.cpp
emveepee: Thanks, done.

phunkyfish:
> > Does it not hide/show depending on the chosen time type?
> 
> Everything is fine. I just don't want it displayed on create. Adding new type with it off just makes the list bigger and still gives this option, I was hoping some sort of tag would help. I probably just need to live with it and am too picky. It is like the menuhook, it gives Activate/Deactivate the Dialog toggle says Enable.

Ya, I say live with it ;)

phunkyfish (approved):

PR #162, opened 2021-02-01
Timeshift fixes

emveepee:
Force radix 10 on stroll, fix missing count on a duration check


emveepee:
@phunkyfish thanks again for the 64 bit cast solution, I am beginning to hate time_t.  Testing by the user on various Android devices confirms that this works.  If there are no other code changes required could you also merge and tag it?

phunkyfish (approved):

PR #164, opened 2021-04-27
Add missing break

emveepee:
Keyword timers where falling through and acting as advanced searches


emveepee:
@phunkyfish @sub3 hopefully this correction to a silly mistake shouldn't take too long to review.

phunkyfish (approved):

PR #165, opened 2021-04-30
Add recording directories

emveepee:
Backend API change allows passing recording directory name on pending shows.

### DIFF
--- a/src/Timers.cpp
+++ b/src/Timers.cpp
@@ -191,8 +191,7 @@ PVR_ERROR Timers::GetTimers(kodi::addon::PVRTimersResultSet& results)
       std::string recordingDirectoryID;
       if (XMLUtils::GetString(pRulesNode, "RecordingDirectoryID", recordingDirectoryID))
       {
-        int i = 0;
-        for (auto it = m_settings.m_recordingDirectories.begin(); it != m_settings.m_recordingDirectories.end(); ++it, i++)
+        for (unsigned int i = 0; i < m_settings.m_recordingDirectories.size(); ++i)
         {
           std::string bracketed = "[" + m_settings.m_recordingDirectories[i] + "]";
           if (bracketed == recordingDirectoryID)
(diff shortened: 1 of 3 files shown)


emveepee:
@ksoo @phunkyfish testing this I noticed that in-progress recordings allow the Recording group and Start padding to be changed.  Is this expected?  The backend does allow the End padding to be changed so an update should be allowed but I was wondering if I could disable these other two options without creating two new timer types

image

src/Timers.cpp
```diff
@@ -319,6 +319,23 @@ bool Timers::UpdatePvrTimer(tinyxml2::XMLNode* pRecordingNode, kodi::addon::PVRT
     }
   }
 
+  if (status == "Pending")
+  {
+    std::string directory;
+    if (XMLUtils::GetString(pRecordingNode, "directory", directory))
+    {
+        int i = 0;
+        for (auto it = m_settings.m_recordingDirectories.begin(); it != m_settings.m_recordingDirectories.end(); ++it, i++)
```
AlwinEsch: As a question of understanding, is there a background to using the iterator and an additional "i" here and on other places?

Could here the following not enough?:
```cpp
  for (size_t i = 0; i < m_settings.m_recordingDirectories.size(); ++i)
```

emveepee: I took existing code https://github.com/kodi-pvr/pvr.nextpvr/blob/Matrix/src/Timers.cpp#L195b for a similar recording group lookup and reused it for this loop.   It dates back to Jarvis, I will change both for readability.

emveepee: @AlwinEsch I made the changes but used unsigned int rather then size_t to avoid compiler warning.

AlwinEsch: No problem, thought only about the size() call, but you are right, the size_t becomes then more in conflict on some other places where then bring warnings.

emveepee:
> 
> @ksoo @phunkyfish testing this I noticed that in-progress recordings allow the Recording group and Start padding to be changed. Is this expected? The backend does allow the End padding to be changed so an update should be allowed but I was wondering if I could disable these other two options without creating two new timer types
> 

I stepped through SetTimer and it seems this is expected behaviour so I created a new timer type which does allow removing these options unfortunately it still gives a user the option of changing the timer type on an in-progress recording which is impossible so I might as just leave it as is.

emveepee:
@phunkyfish  @sub3 although we missed the 19.1 window for packaged binaries, do I need to do anything else for this to be reviewed?

phunkyfish (approved):
Sorry, missed this.

PR #175, opened 2021-07-30
Kodi EDL limit

emveepee:
Kodi PVR limits EDL entries to PVR_ADDON_EDL_LENGTH.  Skip EDL after the limit.

### DIFF
--- a/src/Recordings.cpp
+++ b/src/Recordings.cpp
@@ -602,11 +602,20 @@ PVR_ERROR Recordings::GetRecordingEdl(const kodi::addon::PVRRecording& recording
   tinyxml2::XMLDocument doc;
   if (m_request.DoMethodRequest(request, doc) == tinyxml2::XML_SUCCESS)
   {
-    int index = 0;
+    int numEDL = 0;
     tinyxml2::XMLNode* commercialsNode = doc.RootElement()->FirstChildElement("commercials");
     tinyxml2::XMLNode* pCommercialNode;
     for (pCommercialNode = commercialsNode->FirstChildElement("commercial"); pCommercialNode; pCommercialNode = pCommercialNode->NextSiblingElement())
     {
+      if (numEDL++ == PVR_ADDON_EDL_LENGTH)
+      {
+        /* PVR core stores EDL breaks in a hard-coded array in PVRClient.cpp PVR_EDL_ENTRY edl_array[PVR_ADDON_EDL_LENGTH];
+        * currently https://github.com/xbmc/xbmc/blob/master/xbmc/pvr/addons/PVRClient.cpp#L1056
+        * This check avoids overflowing this arrray */
+         
+        kodi::Log(ADDON_LOG_WARNING, "Maximum EDL entries reached");
+        break;
+      }
       kodi::addon::PVREDLEntry entry;
       std::string buffer;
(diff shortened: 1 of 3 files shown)


emveepee:
@phunkyfish @sub3  because of the long Olympic recordings, a user identified this issue causing a core crash.  Kodi should prevent against the crash in core, but it is also wrong for pvr.nextpvr to try and add beyond the limit.  Hopefully for Olympic viewers this can approved quickly.

src/Recordings.cpp
```diff
@@ -607,6 +607,11 @@ PVR_ERROR Recordings::GetRecordingEdl(const kodi::addon::PVRRecording& recording
     tinyxml2::XMLNode* pCommercialNode;
     for (pCommercialNode = commercialsNode->FirstChildElement("commercial"); pCommercialNode; pCommercialNode = pCommercialNode->NextSiblingElement())
     {
+      if (index++ == PVR_ADDON_EDL_LENGTH)
+      {
```
phunkyfish, line 611: I think this deserves a comment explaining why we should break when the condition is met.

phunkyfish (changes requested):
It’s not obvious we break here

phunkyfish:
> @phunkyfish @sub3 because of the long Olympic recordings, a user identified this issue causing a core crash. Kodi should prevent against the crash in core, but it is also wrong for pvr.nextpvr to try and add beyond the limit. Hopefully for Olympic viewers this can approved quickly.

Yes, agreed. It should be protected.

phunkyfish (approved):

src/Recordings.cpp
emveepee: Ok I added some comments I linked to Kodi master but Matrix has a similar issue.

phunkyfish:
Because we need to move through C we can only have a hard limit. Is there a better limit you would recommend?

emveepee:
> 
> 
> Because we need to move through C we can only have a hard limit. Is there a better limit you would recommend?

I don't really. I figure 10 an hour is a logical maximum, but with 6 hours for our overnight Olympics that would be 60.  

Changing this value probably requires an API bump, at least for me.

phunkyfish:
> > Because we need to move through C we can only have a hard limit. Is there a better limit you would recommend?
> 
> I don't really. I figure 10 an hour is a logical maximum, but with 6 hours for our overnight Olympics that would be 60.
> 
> Changing this value probably requires an API bump, at least for me.

It does. But we can change it once we branch for Nexus. Please create the PR on XBMC master and I’ll include once we branch the addons.

emveepee:
> It does. But we can change it once we branch for Nexus. Please create the PR on XBMC master and I’ll include once we branch the addons.

I may be able to PR to increase PVR_ADDON_EDL_LENGTH to 60 and to change https://github.com/xbmc/xbmc/blob/master/xbmc/addons/kodi-dev-kit/include/kodi/addon-instance/PVR.h#L3130  to avoid the crash.   I am not sure what to do about logging truncated EDL entries in PVR.h

I am uncomfortable with the related GetEpgTagEdl()  https://github.com/xbmc/xbmc/blob/master/xbmc/addons/kodi-dev-kit/include/kodi/addon-instance/PVR.h#L2987 since I don't know how it works and I couldn't find a pvr that SetSupportsEPGEdl(true)

PR #177, opened 2021-08-17
Recording group check

emveepee:
PVRTypeIntValue vectors in GetTimers need to check for empty or size 0 or entries could be re-added when GetTimers () is called more than once.


phunkyfish (approved):
👍

PR #179, opened 2021-08-29
NextPVR API 5.2 changes

emveepee:
NextPVR API 5.2.0 API changes.  Recording size will be forwarded  to core when available.  If the size is not in the XML, use existing logic.   Use radio groups when available.

### DIFF
--- a/src/Channels.cpp
+++ b/src/Channels.cpp
@@ -91,6 +91,11 @@ PVR_ERROR Channels::GetChannels(bool radio, kodi::addon::PVRChannelsResultSet& r
       ++itr;
   }
 
+  if (radio)
+    m_radioGroups.clear();
+  else
+    m_tvGroups.clear();
+
   tinyxml2::XMLDocument doc;
   if (m_request.DoMethodRequest("channel.list&extras=true", doc) == tinyxml2::XML_SUCCESS)
   {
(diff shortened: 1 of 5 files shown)


phunkyfish (approved):
LGTM

src/Channels.cpp
```diff
@@ -228,14 +260,19 @@ PVR_ERROR Channels::GetChannelGroupMembers(const kodi::addon::PVRChannelGroup& g
...
-    for( pChannelNode = channelsNode->FirstChildElement("channel"); pChannelNode; pChannelNode=pChannelNode->NextSiblingElement())
+    for (pChannelNode = channelsNode->FirstChildElement("channel"); pChannelNode; pChannelNode = pChannelNode->NextSiblingElement())
     {
       kodi::addon::PVRChannelGroupMember tag;
-      tag.SetGroupName(group.GetGroupName());
       tag.SetChannelUniqueId(XMLUtils::GetUIntValue(pChannelNode, "id"));
-      tag.SetChannelNumber(XMLUtils::GetUIntValue(pChannelNode, "number"));
-      tag.SetSubChannelNumber(XMLUtils::GetUIntValue(pChannelNode, "minor"));
-      results.Add(tag);
+      // ignore orphan channels in groups
```
phunkyfish, line 267: Can you explain the comment? I'm not following the logic.

emveepee: The comment is because there is a bug in the backend and there could be channels returned that don't exist.  Currently these are logged as errors by core.  

m_channelDetails is existing code and it returns a tuple of  booleans for all channels, first is true when the channel has an EPG, second is true when it is a radio channel.  

The first test is existence

```
m_channelDetails.find(tag.GetChannelUniqueId()) != m_channelDetails.end() 
```

And the second test condition I select when it is true == true (radio) or false==false (TV)

```
group.GetIsRadio() == m_channelDetails[tag.GetChannelUniqueId()].second
```

The backend doesn't support radio or tv groups so I have to do it manually.  There is an assumption that the tv channels must be loaded before the tv group and the radio channel are loaded before radio groups.

Martin

phunkyfish: Ok, cool. Thanks for explaining.

phunkyfish (approved):
👍

emveepee:
@phunkyfish  @sub3 could not tag this 8.2.8-Matrix  has something changed in the tagging process?

phunkyfish:
You mean the GitHub interface change on the release dialog?

It’s a bit unintuitive. You have search for the tag and once it’s not found you get an option to create it.

PR #190, opened 2021-09-18
Changes for settings and sleep

emveepee:
Improve settings for ClientTimeshift in Matrix .  Postpone server check on sleep

### DIFF
--- a/pvr.nextpvr/addon.xml.in
+++ b/pvr.nextpvr/addon.xml.in
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <addon
   id="pvr.nextpvr"
-  version="19.0.0"
+  version="19.0.1"
   name="NextPVR PVR Client"
   provider-name="Graeme Blackley">
   <requires>@ADDON_DEPENDS@
--- a/pvr.nextpvr/changelog.txt
+++ b/pvr.nextpvr/changelog.txt
@@ -1,3 +1,9 @@
+v19.0.1
+- Don't try and reconnect server after sleep announced
+- Adjust timeshift when v4 setting logic get confused
+- Handle situations where Kodi requests timer updates without a recording update
+- Support new API XML to get recording specific art
+
 v19.0.0
 - Translations updates from Weblate
   - be_by, ko_kr, ru_ru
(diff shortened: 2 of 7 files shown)


phunkyfish:
Please submit the PR for Nexus branch first.

emveepee:
Why is that? The Nexus branch will not have the same code and I want to start forking them now, this is not a backport

Martin

phunkyfish:
The Nexus branch was only forked today. So I assume the branches are almost identical.

emveepee:
I started removing the v4 ugly config and also improving sleep in Nexus already when I saw the fork. Before sleep I want to logout the security context which I couldn't do because of V4 users.  It doesn't make sense to me to have to edit and submit PR's for obsolete code This is pre-alpha and there is no guarantee that Nexus works as well as Matrix stable so there is no user issue.  However if the user testing this has no issues I can waste everyone's time if that is how you want it.

AlwinEsch:
About different Version increases and set to related build branches are them a bit different.

You can make a e.g.:
```
git reset --hard HEAD~3
git cherry-pick daf2e747aab29e3f6054cb2ec8ddeef50d448cb1
git cherry-pick da01077d2a989bdb0af0640dc2ecfac59d353dc4
git rebase upstream/Nexus
git push -f
```

And then can be branch here changed to Nexus.

AlwinEsch:
Have changed the branch on here, you need only to call the code from example comment before and then the request correct

emveepee:
I can do the git no problem, I'm working on a branch.  I just don't see the point of doing that and testing it (testing sleep is a nuisance) and then asking @sub3 to merge it when I am redoing it all.  I can submit with no testing but that is not my normal philosophy.

AlwinEsch:
Of course, getting something in without tests would be devastating and absolutely the wrong way to go.

emveepee:
> Have changed the branch on here, you need only to call the code from example comment before and then the request correct

The point of doing this on Matrix was so a user could test it https://forum.kodi.tv/showthread.php?tid=364466&pid=3060825#pid3060825  I will  wait for the user's feedback on deciding what to do

AlwinEsch:
Ah now I understand, yes for this it can be leaved on Matrix.

phunkyfish:
Ah, ok. I see what you are doing now. You could just have explained that in the PR description!

emveepee:
The error is the PDB error and the WIndows Azure builds are fine.   This should be one of the last Matrix builds.  @phunkyfish  and @sub3 can you please review to go with a the new version of the backend.

phunkyfish (approved):
I reran Jenkins. All good 👍

PR #192, opened 2021-10-10
Initial changes for Nexus

emveepee:
Initial support for removal of old versions.  Incorporate two changes from Matrix PR #190

### DIFF
--- a/pvr.nextpvr/addon.xml.in
+++ b/pvr.nextpvr/addon.xml.in
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <addon
   id="pvr.nextpvr"
-  version="20.1.0"
+  version="20.1.1"
   name="NextPVR PVR Client"
   provider-name="Graeme Blackley">
   <requires>@ADDON_DEPENDS@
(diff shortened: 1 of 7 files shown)


emveepee:
@phunkyfish and @sub3 any concerns?  I'd like to get this out soon since I'd like to ensure the block of backend version users is public ASAP.  I also want to start a bigger PR to remove all the other redundant test conditions based on various old versions and remove unnecessary settings.

The two changes from PR#90 have been user tested in a Matrix test build.

phunkyfish:
Ah, you never requested a review 😉

phunkyfish (approved):

phunkyfish:
But just needs addon.xml etc.

emveepee:
> 
> 
> Ah, you never requested a review 😉

Sorry you are always so fast I didn't think to do that.   Do I need to add this? I forgot to ask which is why I didn't update the changelog etc.

```
capabilities.SetSupportsProviders(false);
```

phunkyfish:
> > Ah, you never requested a review 😉
> 
> Sorry you are always so fast I didn't think to do that. Do I need to add this? I forgot to ask which is why I didn't update the changelog etc.
> 
> ```
> capabilities.SetSupportsProviders(false);
> ```

That would be the default anyway. You can add it to be explicit if you  like.

emveepee:
> 
> > ```
> > capabilities.SetSupportsProviders(false);
> > ```
> 
> That would be the default anyway. You can add it to be explicit if you like.

Thanks I will add it.  I also decided to remove the unneeded EDL check in the final commit with the release files.

phunkyfish (approved):

PR #193, opened 2021-10-21
Remove V4 NextPVR support

emveepee:
This PR removes all code and settings needed to support the V4 backend and obsolete versions of V5.


phunkyfish (approved):
Great cleanup!

PR #204, opened 2021-12-28
[Nexus] API related update

AlwinEsch:
Related to xbmc/xbmc#20705 and xbmc/xbmc#20710


emveepee:
@AlwinEsch will you be releasing this or should @sub3  I have been running it as the PR and haven't noticed any problems.

AlwinEsch:
On the morning I continue, by the amount of addons has needed a break time.

PR #208, opened 2022-01-19
Send groups in NextPVR order

emveepee:
Respect channel group delivery sort order from NextPVR.  Can be creation date or  name.  There is no priority in NextPVR but after this change users can create groups in the order they want them displayed in Kodi.

### DIFF
--- a/src/Channels.cpp
+++ b/src/Channels.cpp
@@ -191,19 +191,39 @@ PVR_RECORDING_CHANNEL_TYPE Channels::GetChannelType(unsigned int uid)
 
 PVR_ERROR Channels::GetChannelGroups(bool radio, kodi::addon::PVRChannelGroupsResultSet& results)
 {
-  // radioGroups.size() will be zero before 5.2 API
-  if (radio && m_radioGroups.size() == 0)
+  const std::unordered_set selectedGroups = radio ? m_radioGroups : m_tvGroups;
+  // Many users won't have radio groups
+  if (selectedGroups.size() == 0)
     return PVR_ERROR_NO_ERROR;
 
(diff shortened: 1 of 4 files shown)


src/Channels.cpp
```diff
@@ -191,19 +191,41 @@ PVR_RECORDING_CHANNEL_TYPE Channels::GetChannelType(unsigned int uid)
...
+        }
+      }
+    }
   }
+  else
+  {
+    kodi::Log(ADDON_LOG_DEBUG, "No Channel Group");
+  }
+
+
```
phunkyfish: One blank line should be enough

src/Channels.cpp
```diff
@@ -191,19 +191,41 @@ PVR_RECORDING_CHANNEL_TYPE Channels::GetChannelType(unsigned int uid)
 
 PVR_ERROR Channels::GetChannelGroups(bool radio, kodi::addon::PVRChannelGroupsResultSet& results)
 {
-  // radioGroups.size() will be zero before 5.2 API
-  if (radio && m_radioGroups.size() == 0)
+  const std::unordered_set selectedGroup = radio ? m_radioGroups : m_tvGroups;
```
phunkyfish: `selectedGroup` might not be a good name. Maybe `selectedGroups`?

phunkyfish (changes requested):

src/Channels.cpp
emveepee: Makes sense I was thinking radio or TV but the ,size() indicates groups

phunkyfish (approved):

PR #210, opened 2022-01-26
Remove server-based season episode option.

emveepee:
Kodi Nexus now fully supports season and episodes sorting and display.  Removes now redundant server side option that most users have been using.


phunkyfish (approved):
Woohoo, I love it when we rip code out!

emveepee:
@sub3 could you merge this when you have a chance.  19.4 is out and also supports S/E natively it is time so v20 users should be up to speed now too.

PR #220, opened 2022-11-15
Advanced recording title search

emveepee:
Settings were created to allow a title search feature but it was never implemented

### DIFF
--- a/src/Timers.cpp
+++ b/src/Timers.cpp
@@ -674,8 +674,8 @@ PVR_ERROR Timers::AddTimer(const kodi::addon::PVRTimer& timer)
   switch (timer.GetTimerType())
   {
   case TIMER_ONCE_MANUAL:
-    kodi::Log(ADDON_LOG_DEBUG, "TIMER_ONCE_MANUAL");
     // build one-off recording request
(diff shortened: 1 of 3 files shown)


emveepee:
@phunkyfish when you have chance could you have a look. Thanks.  PR221 is the backport

src/Timers.cpp
```diff
@@ -712,7 +712,7 @@ PVR_ERROR Timers::AddTimer(const kodi::addon::PVRTimer& timer)
 
   case TIMER_REPEATING_EPG:
     if (timer.GetClientChannelUid() == PVR_TIMER_ANY_CHANNEL)
-    // Fake a manual recording not a specific type in NextPVR
+      // Fake a manual recording not a specific type in NextPVR
```
phunkyfish: Move the comment inside the brace if that is what it pertains to. Otherwise, before. 

The two space indent is confusing here.

phunkyfish (changes requested):

src/Timers.cpp
emveepee: @phunkyfish thanks, totally agree, and I re-structured all the comments to make them less confusing.

phunkyfish (approved):

PR #225, opened 2023-01-01
Force group update

emveepee:
Test to see if channel group update loads channels


emveepee:
User confirmed test worked, not required after core PR 22389

PR #236, opened 2023-05-20
Wakeup improvements

emveepee:
After resuming from sleep Kodi will request channel update and if disconnected pvr.nextpvr was not returning an error as required by the API.

Also change to fast poll for servers after wake to account for nic startup and local server startup

### DIFF
--- a/src/Channels.cpp
+++ b/src/Channels.cpp
@@ -88,7 +88,7 @@ PVR_ERROR Channels::GetChannels(bool radio, kodi::addon::PVRChannelsResultSet& r
 {
   if (radio && !m_settings->m_showRadio)
     return PVR_ERROR_NO_ERROR;
-
+  PVR_ERROR returnValue = PVR_ERROR_NO_ERROR;
   std::string stream;
   std::map<int, std::pair<bool, bool>>::iterator  itr = m_channelDetails.begin();
   while (itr != m_channelDetails.end())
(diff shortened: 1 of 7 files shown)


emveepee:
@phunkyfish when you have a chance can you please review.  I can squash them after your review if you prefer.   The Nexus backport is ready.

src/Channels.cpp
```diff
@@ -235,6 +240,9 @@ PVR_ERROR Channels::GetChannelGroups(bool radio, kodi::addon::PVRChannelGroupsRe
       }
     }
   }
+  else {
```
phunkyfish: Brace on the following line

src/Channels.cpp
```diff
@@ -160,9 +160,13 @@ PVR_ERROR Channels::GetChannels(bool radio, kodi::addon::PVRChannelsResultSet& r
       results.Add(tag);
     }
   }
-  return PVR_ERROR_NO_ERROR;
+  else {
```
phunkyfish: Brace on the following line

phunkyfish (approved):
Just some minor brace placement to fix.

src/Channels.cpp
emveepee: Changed thanks

src/Channels.cpp
emveepee: Also changed.

PR #265, opened 2024-06-29
Support content rating 

emveepee:
The pvr.nextpvr addon has never populated the  RarentalRatingCode supplied with the API.  Using XML from  https://movielabs.com/md/ratings/ as a standard for parental rating icons, rating systems and parental rating age value.    The add-on actually uses a parsed version of the for easier processing using a netcore binary that runs on all NextPVR platforms available on request.

The XML file can also use Kodi resource:// link one example provided for the best available parental icon resource source for Kodi https://github.com/wyrm65/resource.images.classificationicons.colour

Users or skins can also create their own resources addons and use a custom XML file in the userdata folder.

In addition to loading icons this XML file is used for populating the ParentalRating numeric field which are not available by API  Although Kodi doesn't use this field to prevent viewing with Parental Controls enabled, a future expansion would make this potentially more important then showing the icon.

Dependent on PR 24096 a new PVR API and move to "P"


emveepee:
@ksooo @phunkyfish is there are way to tag an add-on to not build?  It was uploaded as a draft but that doesn't do it.

phunkyfish:
> @ksooo @phunkyfish is there are way to tag an add-on to not build? It was uploaded as a draft but that doesn't do it.

I don’t believe so. But then again, I’ve never tried before 😉

Galletas78 (approved):

garbear:
Is this still wanted?

emveepee:
> Is this still wanted?

I wrote this when content-ratings were added to Kodi, in case users asked for something that normalizes the data and provided te an image. Each EPG source is different so it can be a real mess.  In any case as I kind of expected no one has actually asked to see the icons some I am happy to have this closed off.

PR #276, opened 2024-12-19
Genre and sub genre mapping

emveepee:
Use the genre mapping xml file to map genre sub types from the English codes send by NextPVR.  Extend genre type and sub type matching to recordings.

### DIFF
--- a/src/utilities/GenreMapper.cpp
+++ b/src/utilities/GenreMapper.cpp
@@ -1,0 +1,182 @@
+/*
+ *  Copyright (C) 2005-2024 Team Kodi (https://kodi.tv)
+ *
+ *  SPDX-License-Identifier: GPL-2.0-or-later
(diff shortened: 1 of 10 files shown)


src/utilities/GenreMapper.cpp
```diff
@@ -0,0 +1,182 @@
...
+
+int GenreMapper::GetGenreType(std::string code)
+{
+  return GetGenreTypeFromCombined(LookupGenreValueInMaps(code));
+};
+
+int GenreMapper::GetGenreSubType(std::string code)
+{
+  return GetGenreSubTypeFromCombined(LookupGenreValueInMaps(code));
+};
```
phunkyfish: Missing newline

phunkyfish (approved):
Just one minor but otherwise good

PR #280, opened 2024-12-24
Trigger channel list updates

emveepee:
Store a hash of the XML channel list and if the hash changes after an EPG update trigger Kodi to reload the channel list and channel groups.  Use cache to remove obsolete channel icons. Add error checking to channel cache logic introduced in Piers.

### DIFF
--- a/src/EPG.cpp
+++ b/src/EPG.cpp
@@ -31,7 +31,10 @@ EPG::EPG(const std::shared_ptr<InstanceSettings>& settings, Request& request, Re
 PVR_ERROR EPG::GetEPGForChannel(int channelUid, time_t start, time_t end, kodi::addon::PVREPGTagsResultSet& results)
 {
   std::pair<bool, bool> channelDetail;
-  channelDetail = m_channels.m_channelDetails[channelUid];
+  {
+    std::lock_guard<std::recursive_mutex> lock(m_channels.m_channelMutex);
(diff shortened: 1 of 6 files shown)


src/EPG.cpp
```diff
@@ -31,7 +31,9 @@ EPG::EPG(const std::shared_ptr<InstanceSettings>& settings, Request& request, Re
 PVR_ERROR EPG::GetEPGForChannel(int channelUid, time_t start, time_t end, kodi::addon::PVREPGTagsResultSet& results)
 {
   std::pair<bool, bool> channelDetail;
+  std::lock_guard<std::recursive_mutex> lock(m_channels.m_channelMutex);
   channelDetail = m_channels.m_channelDetails[channelUid];
+  std::lock_guard<std::recursive_mutex> unlock(m_channels.m_channelMutex);
```
phunkyfish: Just put the protected line in a block instead. You should never need to unlock:

```
  {
    std::lock_guard<std::recursive_mutex> lock(m_channels.m_channelMutex);
    channelDetail = m_channels.m_channelDetails[channelUid];
  }
```

phunkyfish (changes requested):

src/EPG.cpp
emveepee: Thanks I didn't know that was preferred

phunkyfish (approved):

PR #289, opened 2025-01-28
Translation corrections

emveepee:
Added missing an_za and es_es translations that got removed by Weblate.  Correct some en_gb strings

@garbear can you please review, the translations were taken from commit a602bfbc2c696a621b97230c2182a613627252ec


garbear (approved):
Thanks for doing this by hand. I'm glad it's only three files that need to be fixed. Once this is in I'll sync weblate.

PR #311, opened 2025-10-21
Fixup from PR 307

emveepee:
Can't mix upper and lower case translations.  @garbear can you merge this, not sure if it needs @phunkyfish's review.


ksooo:
I'd say keep your hands off those manual Weblate fixups. Should all be fixed directly in weblate.

emveepee:
> I'd say keep your hands off those manual Weblate fixups. Should all be fixed directly in weblate.

OK but en_GB it just the upper case version of en_gb which we are supposed to edit and on Windows they are the same file.  Is there a git setting that will allow me to have both, I want to edit en_gb for my next PR.

emveepee:
Cancelled @garbear while be fixing it via weblate, hopefully soon.

PR #337, opened 2026-06-24
Fix startup hang: always run the connection poll/reconnect thread

kel-mo:
Commit 77d4057922c4fdf0455f8d359805cf17c35653f8 moved connection setup off the main thread but gated the
worker thread's creation on a fully successful session.login: CreateThread()
was only called inside Connect() after both session.initiate and
session.login succeeded.

Because Process() is the only caller of IsUp(), and IsUp() is the only
code path that retries Connect() when the backend is unreachable, a failed
initial connection meant the thread was never started and the connection
was never retried. This is exactly what happens "soon after Kodi startup"
when the NextPVR backend or the network is not ready yet: the client stays
wedged in CONNECTING/SERVER_UNREACHABLE forever and Kodi hangs at
"Creating PVR clients".

The same design also reintroduced a crash: on a transient disconnect the
running worker thread would call IsUp() -> Connect(false) -> CreateThread()
from within its own thread while m_thread != nullptr, which CThread treats
as fatal (logs ADDON_LOG_FATAL and calls exit(1)).

Rework following the pvr.hts model (start the thread from a Start() command,
not the constructor):

- Constructor no longer starts a thread; construction has no side effects.
- Add Start(), which calls CreateThread(). CreateInstance() now constructs
  the client, calls Start(), registers it and returns ADDON_STATUS_OK; it no
  longer connects synchronously on the main thread.
- Process() performs the initial Connect() on the worker thread (so the slow
  backend handshake no longer blocks instance creation), then runs the
  poll/retry loop on !m_threadStop using the interruptible CThread::Sleep()
  so StopThread() returns promptly.
- Connect() establishes the full session in one place (setting.list,
  ReadBackendSettings, ConfigurePostConnectionOptions, SetConnection,
  CONNECTED) and reports failures through the PVR connection state rather than
  ADDON_STATUS_PERMANENT_FAILURE (which CreateInstance no longer consumes
  anyway, now that it registers the instance and returns OK unconditionally).
  An unreachable backend lands in a retryable SERVER_UNREACHABLE state so the
  worker reconnects once the backend becomes reachable again -- e.g. host
  networking/local nextpvr server that is not yet up at kodi startup, or a
  server taken down for maintenance. A rejected login (ACCESS_DENIED) or a
  too-old/incompatible backend (VERSION_MISMATCH) is reported with its
  specific reason and left sticky -- IsUp() does not poll those states -- so
  the user is informed without hammering a known-bad credential or
  re-notifying on every attempt; the instance stays alive and recovers on the
  next wake, settings change, or restart. Connect() no longer touches threads,
  removing the exit(1).
- Use StopThread() in the destructor; the previous StopThread(2000) passed
  2000 as the bool 'wait' argument, not a timeout.
- Remove the duplicate std::atomic<bool> m_running, which shadowed and
  collided with CThread's own private m_running member.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

### DIFF
--- a/src/addon.cpp
+++ b/src/addon.cpp
@@ -50,14 +50,12 @@ ADDON_STATUS CNextPVRAddon::CreateInstance(const kodi::addon::IInstanceInfo& ins
       client = new cPVRClientNextPVR(instance);
     }
 
-    status = client->Connect();
-
-    if (status != ADDON_STATUS_PERMANENT_FAILURE)
-    {
-      status = ADDON_STATUS_OK;
-      hdl = client;
-      m_usedInstances.emplace(std::make_pair(instance.GetID(), client));
-    }
+    // The client connects and maintains the connection on its own worker
+    // thread; connection problems are reported via the PVR connection state.
+    client->Start();
+    hdl = client;
+    m_usedInstances.emplace(std::make_pair(instance.GetID(), client));
+    status = ADDON_STATUS_OK;
   }
 
   return status;
--- a/src/pvrclient-nextpvr.cpp
+++ b/src/pvrclient-nextpvr.cpp
@@ -113,7 +113,6 @@ cPVRClientNextPVR::cPVRClientNextPVR(const kodi::addon::IInstanceInfo &instance)
   m_realTimeBuffer = new timeshift::DummyBuffer(m_settings, m_request);
   m_livePlayer = nullptr;
   m_nowPlaying = NotPlaying;
-  m_running = true;
 }
 
 cPVRClientNextPVR::~cPVRClientNextPVR()
@@ -135,8 +134,7 @@ cPVRClientNextPVR::~cPVRClientNextPVR()
       CloseLiveStream();
   }
 
-  m_running = false;
-  CThread::StopThread(2000);
+  StopThread();
   kodi::Log(ADDON_LOG_DEBUG, "->~cPVRClientNextPVR()");
   if (m_bConnected)
     Disconnect();
@@ -145,6 +143,14 @@ cPVRClientNextPVR::~cPVRClientNextPVR()
   m_recordings.m_hostFilenames.clear();
   m_channels.m_channelDetails.clear();
   m_channels.m_liveStreams.clear();
+}
+
+void cPVRClientNextPVR::Start()
+{
+  // Launch the worker thread, which performs the initial connection and then
+  // keeps polling/reconnecting. The thread is started here rather than in the
+  // constructor so construction has no side effects.
+  CreateThread();
 }
 
 ADDON_STATUS cPVRClientNextPVR::Connect(bool sendWOL)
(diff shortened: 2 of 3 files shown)


emveepee:
There are some good things there, but I don't want to start a thread in the constructor  Check how pvr.hts starts the threads after connection with the Start command.

kel-mo:
> There are some good things there, but I don't want to start a thread in the constructor Check how pvr.hts starts the threads after connection with the Start command.

Fair enough and thanks for the feedback. I've adapted accordingly and force pushed a run tested re-work.

emveepee:
I haven't tested this but I don't see a permanent failure possibility.  Versions, localhost and Android come to mine.

kel-mo:
> I haven't tested this but I don't see a permanent failure possibility. Versions, localhost and Android come to mine.

Many failures are transient and recoverable, so long as the reason is surfaced to the user for their action. Please do test. Current pvr.nextpvr Piers branch head is not working at all to recover from errors at startup, and then after intermittent failure in perfect conditions gets compounded by a bug that prevents a timely and clean shutdown.

emveepee:
I agree that changes are required.  Some errors are permanent such as an initial install that fail.  Please ensure that there is a permanent failure  on the version mismatch, if a connect has never occurred, invalid PIN etc.

emveepee:
@kel-mo are you making any more changes or are you ready for this to be converted from draft and get tested?

kel-mo:
> @kel-mo are you making any more changes or are you ready for this to be converted from draft and get tested?

I attempted to re-introduce ADDON_STATUS_PERMANENT_FAILURE state as part of the initial configuration/instance onboarding and found that it added much complexity for little perceived benefit. It also resulted in a worse user experience, for example the wrong PIN? notification would quickly be replaced with a notification that the addon failed to start. The worker thread can be tuned to not continue polling a problematic configuration while the admin remediates the issue surfaced to them and restarts/disables the addon.

kel-mo:
Also, if there were more than one server configuration, and if there was an external error preventing the second configuration from succeeding but the first one was fine, wouldn't declaring ADDON_STATUS_PERMANENT_FAILURE kill both? I feel like that status should be reserved for the most fatal errors afflicting the addon itself, and that the addon worker thread should just be reporting PVR_CONNECTION_* status when the addon is functional and able to do so.

That said, I am not all that familiar with the code and defer to you better judgment. Mine is clouded by a crappy virus and might be compromised!

emveepee:
I disagree a wrong PIN is permanent failure.   A bad configuration on any server instance needs the failure so the user knows to go settings and make a change otherwise won't they think it is working, and then it becomes a support issue.  In fact with multiple servers it might be more important since the user might not notice that one server is not responding.

kel-mo:
> I disagree a wrong PIN is permanent failure. A bad configuration on any server instance needs the failure so the user knows to go settings and make a change otherwise won't they think it is working, and then it becomes a support issue. In fact with multiple servers it might be more important since the user might not notice that one server is not responding.

If the worker thread is kept alive it can periodically remind the end user via notifications that something is wrong with their configuration. If we declare total addon bad because of one config, the end user only gets 1 fleeting message that's covered up, blink and you miss it.

emveepee:
It isn't "one" config it is NextPVR is an an insane loop that can never be corrected without additional user action.  I am not sure why you don't want to simply make the addon work the way it always did in the past, Historically pvr.nextpvr  was based on the MeadiaPortal addon which continues to set the same flag and I haven't read any complaints about either addon's approach.

I am used to that behavior and it is the only way II want to test it but feel free to submit the PR and let TK and @sub3 make the decision if you think it is an improvement.

kel-mo:
Thanks for the feedback, I won't be progressing this further.

